PR #23, opened 2026-07-06
Repository layout and development environment

leonfullxr:
Closes #12

# Description

Establishes the monorepo structure and a reproducible development environment, implementing the layout specified in the documentation from #22.

# Details

- frontend: SvelteKit skeleton with adapter-static and an SPA fallback (one build artifact for every deployment), prettier and svelte-check wired as the lint and check scripts.
- backend: FastAPI skeleton serving GET /api/v1/health (process state only) and GET /api/v1/config (auth_methods, billing_enabled, languages), with pydantic-settings reading AUTH_MODE, OAUTH_PROVIDERS and BILLING_ENABLED. Tests cover both endpoints and the auth method derivation per mode.
- worker: installable Python package with the DEVICE setting (cuda, rocm, cpu) and a runnable module entry point. Inference dependencies deliberately arrive with issue #15 so this skeleton installs anywhere.
- deploy/compose/dev.yml: PostgreSQL, Redis, MinIO and Mailpit for native development. MinIO and Mailpit host ports are offset from their defaults because development machines often run other services on them.
- Replaced the Java template .gitignore with entries for Python, Node and project artifacts.
- docs/local-development.md gains a "Running each component" section; the README gains a short development pointer.

One deliberate adjustment to the issue text: the issue asked for a development compose file starting the API server, worker and PostgreSQL. The development loop decision in docs/decisions.md (dependencies in containers, applications native) supersedes that, so the dev compose starts dependencies only; the full self-hosted stack is issue #18's deliverable.

# Verification

All run on the reference development machine: backend ruff and pytest green, uvicorn smoke tested (health and config return the expected payloads, oauth mode derives ["local","google","github"]); worker ruff and pytest green, python -m worker runs; frontend npm run lint, check and build green with the static adapter writing build/; the compose stack starts all four services (verified healthy over HTTP) and tears down cleanly.

Draft until #22 merges; based on initial-docs so it retargets to main automatically when #22 lands.


leonfullxr:
Review notes on this pull request are addressed higher in the stack, in #25 commit 7372d61 (and 819eaf7): oauth_providers parsing now strips whitespace with a regression test, the minio and mailpit images are pinned, and the nonstandard text-scale meta tag is removed. Fixing them here would force a rebase cascade through #24 and #25 for no change in the merged result.

PR #24, opened 2026-07-06
Continuous integration

leonfullxr:
Closes #13

# Description

Adds the GitHub Actions workflow running lint and unit tests for all three components on every pull request, implementing stage 1 of the pipeline specified in docs/local-development.md from #22.

# Details

- frontend job: npm ci, prettier check, svelte-check, production build with the static adapter.
- backend job: pip install, ruff, pytest on Python 3.11 (matching the pyproject minimum).
- worker job: same shape as the backend.
- Later pipeline stages are deliberately out of scope here and arrive with their issues: the tiny model CPU integration test with issue #15, backend integration against service containers with issue #16, image builds to GHCR and the cloud-sim smoke test with issue #18.

# Verification

The workflow YAML parses and each job runs exactly the commands verified locally on the repo-layout branch (all green there). The jobs execute for real on this pull request the moment it is opened, which is the true test.

Stacked on the repo-layout branch (PR #23), so it retargets automatically as the chain merges: #22, then #23, then this.


leonfullxr:
Review notes on this pull request are addressed higher in the stack, in #25 commit 7372d61: the workflow is split into one per component (frontend, backend, worker, plus the cross-component simulation), each triggered only when its files change as issue #13 specifies, with contents read permissions and python -m pip. One consequence to remember when enabling branch protection: path-filtered workflows do not run on unrelated changes, so required checks must be chosen accordingly.

PR #99, opened 2026-07-16
Pipeline phase timing for job bottleneck metrics

leonfullxr:
## Summary
- Workers time input fetch, model load, denoise (`gpu_ms`), and post-process (encode/upload) phases and send optional spans on `job_done`.
- Migration 0004 adds `input_fetch_ms`, `load_ms`, `postprocess_ms`, and `failure_reason` on job rows; the generations API exposes them with `dispatched_at` / `finished_at`.
- Studio usage metrics show queue wait p50/p95, post-process share, effective GPU utilization, and session failure rate.

Closes #96

## Test plan
- [x] `make verify` locally
- [x] Self-hosted CI (start runner with `make ci-runner-start` if idle)
- [x] `make simulate` - generate an image and confirm phase fields appear in `/api/v1/generations`
- [ ] Open Studio metrics Usage tab - pipeline cards populate after a few jobs


leonfullxr:
Review: the shape is right (optional wire fields are N-1 safe, migration 0004 is clean both ways, tests cover success and failure paths, failure_reason persisting the stall/OOM text is a nice operator win). Three findings before this leaves draft:

1. Bug - the failure-rate tile always reads 0. loadHistory strips failed generations at fetch (withoutFailed at studio.svelte.ts:138), so computePipelineMetrics(studio.history) never sees state failed and failureRatePct is 0 percent for any session with failures. Either keep failed rows in studio.history and filter at the display sites that need it (the strip and gallery already filter on their own), or drop the tile until history carries failed jobs. Do not ship a metric that cannot move.

2. Metric definition - effectiveUtilPct sums per-job walls (dispatched to finished), which excludes idle BETWEEN jobs. That is a useful in-job GPU share, but it is not the docs' effective-utilization KPI (gpu_ms / worker wall clock), and it will understate the win from #97's depth-2 dispatch, whose whole point is removing between-job idle. Suggest: rename the tile to reflect in-job share, and add a span-based session utilization (total gpu_ms / (max finished_at - min dispatched_at)) when the session has two or more terminal jobs - that is the number #97's acceptance criterion needs.

3. Seam leak - run_job special-cases isinstance(engine, SimulatedEngine) to account for model load. The engine seam should stay opaque: have SimulatedEngine.generate mark itself loaded and report the simulated load in GeneratedImage.load_ms like DiffusersEngine does, and delete the isinstance branch.

Minor: es.json 'Utilizacion GPU' is missing its accent (Utilización, the file uses accents elsewhere); _generate reuses the start variable for load and gpu timing while _generate_i2i uses a separate load_start - pick the second style in both for readability.

PR #100, opened 2026-07-16
Pipeline overlap: dispatch depth 2 per worker

leonfullxr:
## Summary
- Replace `job_busy` boolean with bounded `jobs_in_flight` counter (limit 2) on each worker.
- `pick_job_worker` selects workers below the depth cap so job N+1 dispatches while job N encodes/uploads.
- Worker and wire protocol unchanged; per-job inflight bookkeeping, watchdog, and retry paths unchanged.

Closes #97

## Test plan
- [ ] `make simulate` - post 10 turbo jobs; Session GPU utilization tile should reach >=90% (was ~55-65%)
- [ ] Confirm realtime slot accounting unaffected

### DIFF
--- a/backend/app/jobs.py
+++ b/backend/app/jobs.py
@@ -273,11 +273,21 @@ async def generation_events(
     )
 
 
+def job_dispatch_depth(worker: realtime.Worker) -> int:
+    # Sessions-first: while a realtime slot is live, do not stack a second
+    # queued job behind the one already waiting on the GPU lock.
+    if worker.slots_in_use > 0:
+        return 1
+    return JOB_DISPATCH_DEPTH
+
+
 def pick_job_worker(model_id: str) -> realtime.Worker | None:
-    for worker in realtime.workers.values():
-        if model_id in worker.models and worker.jobs_in_flight < JOB_DISPATCH_DEPTH:
-            return worker
-    return None
+    candidates = [
+        worker for worker in realtime.workers.values()
+        if model_id in worker.models
+        and worker.jobs_in_flight < job_dispatch_depth(worker)
+    ]
+    return min(candidates, key=lambda worker: worker.jobs_in_flight, default=None)
 
 
 def release_job_slot(worker: realtime.Worker) -> None:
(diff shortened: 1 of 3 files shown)


leonfullxr:
Review: the mechanics match the #97 design - bounded counter with a floor guard in release_job_slot, all five release sites converted, no wire or worker changes, and the depth tests cover both the overlap and the cap. Two findings before merge:

1. Multi-worker packing: pick_job_worker returns the FIRST worker under the depth limit, so with two workers connected it stacks two jobs on worker A while worker B sits idle - worse latency than depth 1 in that topology. Pick the least-loaded candidate instead (min by jobs_in_flight among workers that have the model and are under the cap), mirroring how pick_worker chooses sessions by max free slots. One line, and it belongs in this PR because this PR introduces the packing behavior.

2. Session contention doubles: a realtime frame now queues behind up to TWO jobs' denoise on the worker GPU lock (worst case 2x20s behind sdxl-base class, previously 1x). The recorded scheduling principle is that queued jobs fill idle capacity and sessions come first - suggest honoring it cheaply: treat the effective depth as 1 for workers with slots_in_use > 0, so a drawing user never pays the doubled wait. Alternatively document the tradeoff, but the guard is three lines and matches the decision on record.

The acceptance checkbox (10 turbo jobs, session utilization >= 90) still needs the real-GPU run - measuring it now on the RX 7600 XT and will post the numbers.

leonfullxr:
Acceptance measured on the RX 7600 XT (clean GPU, isolated stack, vega-rt at 512/2, one warmup job then ten measured, session utilization = sum gpu_ms / (max finished_at - min dispatched_at)):

| Revision | Succeeded | gpu_ms total | Span | Session utilization |
| --- | --- | ---: | ---: | ---: |
| main (depth 1) | 10/10 | 3617 | 4909 ms | 73.7 percent |
| this PR (depth 2) | 10/10 | 3530 | 3831 ms | 92.1 percent |

The acceptance criterion (>= 90 percent) passes, and the same batch completes 28 percent faster end to end. Post-process time (372 ms total) now overlaps the next job's denoise instead of holding the slot. Both review findings verified in the code: least-loaded packing and the sessions-first depth guard.

PR #104, opened 2026-07-17
Add Real-ESRGAN pixel upscale path on the worker

leonfullxr:
## Summary
- Add `upscale` capability with load-time exclusivity against diffusion capabilities (issue #89 / design on #90).
- Ship `realesrgan` manifest (factor enum 2/4, BSD-3-Clause weights) and tiled spandrel inference so 1024 x4 stays VRAM-bounded.
- SimulatedEngine Lanczos path for CI; tile math + exclusivity unit tests.

## Test plan
- [x] `make verify` locally
- [ ] CI green on self-hosted runner
- [ ] Manual (before merge if GPU free): measure VRAM/latency for 512x2, 1024x2, 1024x4 on RX 7600; confirm `min_vram_gb` (provisional 4 GB)
- [ ] Confirm weights download once into `models/.cache/upscale/`

Closes #89


leonfullxr:
Review: faithful to the design - spandrel (not realesrgan/basicsr), tiled from day one with feathered blending and normalization, capability exclusivity enforced at load with a test, correct release URLs and BSD-3 license recorded, SimulatedEngine path keeps CI GPU-free.

Two fixes applied on the branch:

1. Bug (a3f94de): the blend ramp started at exactly zero, so the outermost pixel row of every tile carried zero weight. Interior seams were covered by neighboring tiles, but at image borders only one tile contributes and 0/(0+1e-8) divides to black - a one pixel dark frame around every upscaled image. The ramp now starts one step in (normalization reconstructs single-coverage pixels exactly for any positive weight) and a regression test runs the real tiled path over a white 700x520 image (2x2 tiles) asserting nothing darkens.
2. Install-order footgun (410ea15): spandrel pulls torchvision, and resolving torchvision from PyPI drags in the CUDA torch build, silently replacing the ROCm wheel (observed live: 2.9.1+rocm6.3 became 2.13.0+cu130 and the worker died with 'Found no NVIDIA driver'). make setup-rocm now installs torch AND torchvision from the ROCm index before the extra.

End-to-end on the RX 7600 XT (full stack, vega-rt 1024px source): x2 -> 2048x2048 in 17.9 s gpu_ms, x4 -> 4096x4096 in 37.9 s, VRAM bounded by tiling throughout, border samples healthy against the interior mean, phase fields populated. Weights download once (~65 MB per factor) and cache under models_dir.

Optional follow-up, not this PR: RRDBNet is the heavy Real-ESRGAN variant; the compact realesr-general-x4v3 net is several times faster at slightly lower fidelity and would drop the x2 wait under ten seconds - worth a second manifest if upscale latency matters.

leonfullxr:
Copilot triage, fixes in 7c468dd:

- ensure_weights: downloads now go through urlopen with a 60 s socket timeout into a per-process temp file (workers sharing one models_dir cannot clobber each other), atomic replace, temp cleaned up on failure.
- weight_cache_path validates the factor and raises the same ValueError as weight_url instead of a bare KeyError.
- Rung decision is no longer decorative for upscalers: _select_rung pins them to full residency (offload rungs are diffusers pipeline mechanics; these are plain nn.Modules run tiled), and _load_upscale drops its ignored _pick_rung call.
- Preload reads the manifest's declared factor default instead of hardcoding upscale-2.

Declined:
- torch.frombuffer in _pil_to_tensor: the conversion is milliseconds against 0.5-38 s of convolution; not worth the version-gated fallback.
- _tile_starts redundant starts: deliberate simplicity; the unique pass is the documented invariant and rewriting the generator risks coverage edge cases for zero measured benefit.
- IndexError for overlap >= length: unreachable with the module constants; the arguments are programmer-controlled and a crash is the diagnostic.

PR #105, opened 2026-07-17
Add upscale job gates on POST /generations

leonfullxr:
## Summary
- Extend `POST /api/v1/generations` for capability-driven upscale (source required; diffusion or upscale for sourced jobs).
- Mirror upscale/diffusion exclusivity on fleet hello; factor-aware estimates + provisional `realesrgan` timing.
- Document pixel upscale vs refine-as-i2i and the dedicated `MODELS_DIR` worker pool pattern.

Stacked on #104 (issue #89). Retarget base to `main` after that merges.

## Test plan
- [x] `make verify` locally
- [ ] CI green (DB tests cover upscale dispatch / rejects / exclusivity)
- [ ] After #104 merges: retarget this PR to main

Closes #90


leonfullxr:
Review: every gate from the design review is implemented and tested - source required for upscale models, the source gate accepts either capability, exclusivity rejected at hello with a wire-level test, the refine tier documented as the existing i2i path, factor-keyed estimates fail closed. The provisional 800 ms baseline is replaced with the measured number (7c9ee32): x2 on a 1024px source is 17931 ms on the reference card. Note in the source comment: x4 measures 37853 ms, so quadratic factor scaling overestimates it by about two - a per-factor map is a reasonable future refinement if the estimate label bothers anyone.

PR #106, opened 2026-07-17
Enable studio Upscale with factor picker

leonfullxr:
## Summary
- Enable Upscale on finished history images with x2/x4 factor toggle.
- Pin the live `upscale` model; hide it from the diffusion model select.
- Show factor-scaled GPU estimates from the flat realesrgan baseline.

Stacked on #105 (issue #90). Retarget toward main after #104 and #105 merge.

## Test plan
- [x] `make verify` locally
- [ ] CI green
- [ ] Manual: generate at 512/1024, Upscale x2 and x4, confirm history + estimate label
- [ ] Confirm Upscale stays disabled when no upscale model is registered

Closes #91


leonfullxr:
Copilot triage, fixes in dd852cf:

- Real bug, good catch: upscaleShown cleared studio.selectedId before reading shown, and shown is derived from selectedId, so upscaling an older selected image targeted the newest one. The asset and model ids are now captured before the selection is cleared.
- The diffusion filter is centralized as filterDiffusionModels in studio.svelte.ts and used by the generate form, applyModels, and the sidebar model list, which previously could still select the upscaler into studio.modelId.
- applyModels no longer falls back to upscale models: with no diffusion models registered it clears modelId and the form shows the no-models state while the Upscale action stays available.

Deferred: preferring a default-flagged upscaler applies once a second upscaler exists; that is issue #108 territory and the picking policy will be decided there.

PR #113, opened 2026-07-17
Add realesrgan-fast compact upscale manifest

leonfullxr:
## Summary
- Ship `realesrgan-fast` (realesr-general-x4v3, BSD-3) alongside quality `realesrgan`.
- Direct `.pth` manifest sources use one weight for every factor; requested factor below native scale runs at native then Lanczos-downsamples (RealESRGANer outscale).
- Flat timing baseline 550 ms (measured x4 at 1024 on RX 7600 XT); studio still selects quality upscaler first by sorted id.

## Test plan
- [x] `make verify` locally
- [x] CI green
- [x] Manual: register both upscalers; Upscale x2/x4 with `realesrgan-fast` via API; confirm sub-second on 1024 and no dark border
  - Confirmed on RX 7600 XT via branch path (019cfb1): x2 2048 in 759 ms, x4 4096 in 555 ms, peak 904 MB, single 4.7 MB weight, healthy borders (review comment on this PR).
- [x] Confirm studio Upscale still pins `realesrgan` (sorted id) until a deliberate default decision
  - `GET /api/v1/models` sorts by id; `generate-panel` uses `.find(upscale)` so `realesrgan` wins over `realesrgan-fast`.

Closes #108


leonfullxr:
Reviewed and GPU-verified. The implementation matches the #108 design: direct .pth source for all factors, native scale from the spandrel descriptor via UpscaleRuntime, Lanczos downsample when factor is below native (RealESRGANer outscale behavior), flat estimate entry, licensing row, and tests covering both the downsample path and the factor-above-native rejection.

One fix pushed (019cfb1): the branch reintroduced the no-op self._pick_rung call in _load_upscale that was removed during the #104 review round; every caller already populates the rung and upscalers are pinned to full residency in _select_rung.

Acceptance on the RX 7600 XT through the branch's real path (load_upscale_model + upscale_tiled with native_scale=4, 1024px source, fp16, MIOpen warmup):

- x2: 2048x2048 in 759 ms (native x4 run plus CPU-side Lanczos down)
- x4: 4096x4096 in 555 ms
- peak VRAM 904 MB, weights a single 4.7 MB file shared by both factors, borders healthy

Worker suite 45 passed, ruff and mypy clean. Notes: the flat 550 ms estimate slightly undershoots x2 (resize overhead lands it near 750 ms) - both read as under a second in the UI, so not worth a second timing knob. Model ordering keeps realesrgan (quality) as the studio Upscale button's pick since the registry sorts by id; making fast the default is the side-by-side product call noted in #108 and stays open.

PR #114, opened 2026-07-17
Evict poisoned pipelines after generation errors

leonfullxr:
## Summary
- After a non-OOM generation or frame error, drop the resident model so the next job reloads clean (OOM still uses the existing evict-and-retry path).
- Cooldown (30s) and eviction counts avoid silent load/unload thrash; ValueError/TypeError stay as request errors without eviction.
- Regression coverage: fail once, succeed after reload; cooldown; validation errors do not evict.

Closes #103.

## Test plan
- [x] `make verify`
- [x] CI green on self-hosted runner
- [x] Optional live check: covered by unit regression reproducing the c10::Half signature from the #102 incident (review: no code changes needed)

### DIFF
--- a/worker/tests/test_engine.py
+++ b/worker/tests/test_engine.py
@@ -156,14 +156,17 @@ def test_evict_cold_removes_oldest_first():
 
 
 def _poison_engine(model_id: str = "m") -> DiffusersEngine:
-    import torch
-
     from worker.engine import GeneratedImage
 
+    # CI worker venv has no torch; only OutOfMemoryError must be a real type
+    # so `except self.torch.OutOfMemoryError` stays valid.
+    torch_stub = MagicMock()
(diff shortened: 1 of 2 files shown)


leonfullxr:
Reviewed, no changes needed. The exception classification is right: ValueError/TypeError re-raise without eviction (request validation - the upscale factor gate and the realtime checks raise these, and a bad request must not evict the resident), OutOfMemoryError keeps its existing retry path, and everything else drops the resident with a 30 s cooldown so a permanently broken weight set fails fast instead of thrashing load/unload. The tests reproduce the actual c10::Half failure signature from the #102 incident.

Two observations, neither blocking: the import logging line lands after time instead of alphabetically; and when the OOM-retry second attempt raises a poison-class error it propagates without eviction - the next job's failure will evict then, which is acceptable self-healing for a double-failure path.

PR #119, opened 2026-07-17
Studio Upscale peer panel and history drag-scroll

leonfullxr:
## Summary
- Make Upscale a peer left-panel mode beside Generate (shared viewer and history strip), with Fast (`realesrgan-fast`) default and Quality (`realesrgan`) as an option.
- Keep upscalers out of the diffusion model picker; empty states cover missing finished images or unregistered upscale models.
- Add click-hold drag horizontal scrolling on the history thumbnail strip (scrollbar remains hidden).

## Test plan
- [ ] Open studio with a worker that announces both `realesrgan-fast` and `realesrgan`.
- [ ] Toggle Generate | Upscale; confirm Generate still shows diffusion controls and Upscale shows Fast/Quality + x2/x4.
- [ ] Upscale a finished history image with Fast and Quality; confirm jobs appear in history and viewer updates.
- [ ] With Fast unavailable (worker without `realesrgan-fast`), confirm Quality remains selectable and Fast is disabled.
- [ ] With many history thumbnails, click-hold and drag left/right on the strip to scroll; short clicks still select.
- [ ] `make verify` passes (already run locally).


leonfullxr:
Reviewed all three commits. The panel architecture is right: Upscale as a peer mode sharing the viewer and history, fast default via defaultUpscaleModelId with quality as an explicit tier, upscalers kept out of the diffusion picker and sidebar, i18n complete in both languages, docs updated, and the capture-before-clear fix from #106 preserved in upscaleShown. The drag-scroll evolution is sound: capturing only after the 12 px threshold (2b323f6) is what keeps plain clicks selecting, and the suppressClick flag resets on the next pointerdown so no clicks are swallowed after a drag. The fling (d97989c) handles the details that usually go wrong: EMA-smoothed velocity, zeroed when the pointer sits still 80 ms before release, dt-clamped exponential decay, bounds clamping with stop at the edges, and cancellation on grab and destroy.

One gap this PR exposes, fixed separately in #120: the estimate label used factor-squared scaling, which shows ~2.2 s for a fast x4 that takes 0.55 s (the compact net runs at native x4 regardless of factor) and 71.7 s for a quality x4 measured at 37.9 s. #120 wires measured per-factor numbers through GET /api/v1/models and makes estimateUpscaleGpuMs prefer them; it is independent of this branch and merges in either order.

Non-blocking note: the JS fling ignores prefers-reduced-motion (the CSS tile animations respect it); a one-line matchMedia guard in startFling would cover it.

PR #214, opened 2026-08-04
Decode realtime frames with a distilled VAE

leonfullxr:
# Description

Closes #196.

A manifest can name a distilled decoder used **only** on the realtime frame path. `vega-rt` names `madebyollin/taesdxl`.

# Motivation

Profiling the frame before reaching for the published techniques changed the target. Of a 265.8 ms frame on the reference card:

| Stage | Median | Share |
| --- | --- | --- |
| UNet plus scheduler | 140.8 ms | 53 percent |
| **VAE decode** | **122.2 ms** | **46 percent** |
| Output conversion | 2.8 ms | 1 percent |

Stream Batch and residual classifier-free guidance both address the other half. Nearly half the frame was the decoder, and `vega-rt` already runs the fp16-safe VAE in fp16, so that 122 ms is real compute rather than an fp32 upcast.

# Functionality

`madebyollin/taesdxl` is 2.45M parameters against the SDXL VAE's roughly 84M, MIT licensed, and its model card describes it as being for real-time previewing of SDXL generation. Measured on this card:

| | Before | After |
| --- | --- | --- |
| VAE decode | 123.7 ms | **8.3 ms** |
| End-to-end `frame()` | about 274 ms | **184.9 ms** |
| **Calibrated slots** | **1** | **3** |

Roughly a threefold reduction in cost per active session, against a target of doubling.

**Only `frame()` uses it.** `vega-rt` also serves `text_to_image` and `image_to_image`, which produce the archival PNG masters under the stored-outputs decision, and those keep the full VAE. Silently degrading saved work would be far worse than a slow preview.

The field is worker-side only and excluded from `wire()` like `source`, `vae`, `scheduler`, `lora` and `quantize`. The decoder loads once, lives on its pipeline so a model switch drops it, and a deterministic load failure falls back to the full VAE with a warning rather than breaking realtime.

# Details

**Three things only real hardware found, all the same shape: the pipeline used to do it for us.**

1. `.images[0]` dropped the batch dimension. 85 tests were green and the feature did not work at all, because the fake returns a list so `[0]` is well defined for it.
2. The decode needed an inference context. Diffusers wraps its own decode internally; calling the decoder directly left autograd on and `postprocess` refused the tensor.
3. Review found that an out-of-memory load was cached as permanent absence. Beyond blocking the existing evict-and-retry path, it let capacity diverge from reality: frames would fall back to the roughly 274 ms full-VAE path while the engine kept advertising the three slots it calibrated at 182 ms, admitting three sessions to a card that could hold one. Out-of-memory now propagates, only deterministic failures are cached, and falling back clears the calibrated slot count so the next advertisement measures the path that will actually run.

**The tests now model the real contracts.** The fake returns a `(1, 4, 64, 64)` tensor and asserts four dimensions, a test asserts the inference context, and a numerical test decodes a fixed latent on the CPU and asserts the output is finite, in range and not degenerate. That last one skips unless torch, diffusers and cached weights are all present, so CI stays green; it passes in 8 seconds where they are.

# Verification

`make verify-worker`: ruff, mypy, **87 passed, 5 skipped**.

On hardware, through the real `frame()` path: a correct 512 px image, median 184.9 ms end to end, three calibrated slots. I viewed the output rather than only checking statistics.

# Review status

**Not reviewed independently.** Copilot is over quota and the adversarial reviewer hit its usage limit before the second pass. Its first pass produced the two findings above, which are fixed, and I checked the fixes myself against the two non-obvious properties: that the out-of-memory clause precedes the generic handler, which matters because `OutOfMemoryError` subclasses `RuntimeError` and `RuntimeError` is in the deterministic list, and that `_calibrated_slots` is already `int | None` with `None` meaning not yet calibrated, so clearing it is the existing invalidation path rather than a new state.

That is my own review of work I delegated, not an independent one. Left as a draft for that reason.


leonfullxr:
Two notes from measuring realtime concurrency on the reference card, both of which raise this PR's priority.

**This is the thing that delivers two concurrent users.** Cross-session batching was expected to be the mechanism and it is not: measured at 512 px with fused attention, vega-rt at two steps costs 264 ms of complete frame for one session and 465 ms for two batched, which leaves 35 ms against the 500 ms bar before the 30 to 50 ms collection window a batching scheduler needs. Your decode substitution is far bigger: 122.2 ms of a 265.8 ms frame against 8.3 ms for TAESDXL, and decode is per image, so a batch pays it per member. That puts one session near 150 ms, two serialised near 300 ms and a batch of two near 238 ms. Two users then need no scheduler change at all.

**One inconsistency to resolve before merging.** The claims of a 184.9 ms end-to-end median and three calibrated slots cannot both hold. Slots are floor(500 / p95) and p95 is never below the median, so 184.9 ms supports two slots at most; three requires p95 at or below 166.67 ms. Worth reporting the p95 rather than the median, since the formula reads the p95 and a review will check the arithmetic.

leonfullxr:
Measured this independently of the branch, because it decides whether concurrency needs a scheduler at all, and it holds. Same card, `vega-rt` at two steps, 512 px, 60 timed frames per point after a discarded warmup, fused attention on, complete frame including the WebP encode. The decoder is the one this PR names, called the way this PR calls it: `output_type="latent"` then `decode`, with no scaling divide.

| sessions in one call | full VAE p95 | distilled p95 |
|---|---|---|
| 1 | 267.8 ms | **148.8 ms** |
| 2 | 465.5 ms | **226.1 ms** |
| 3 | 675.0 ms, past the bar | **307.3 ms** |

Serialised, that is 298 ms for two sessions and 447 ms for three, both inside the 500 ms bar with no batching and no scheduler change. So this PR is what delivers concurrent users, and cross-session batching (#294) becomes the mechanism for the fourth and beyond rather than the second. I had it the other way round and said so on #294.

Your profile is confirmed as the reason: 122.2 ms of decode in a 265.8 ms frame, and decode is per image, so a batch pays it per member. That is why the saving compounds with concurrency while the seventeen percent that batching buys does not.

Quality, honestly bounded: I rendered the same canvas, prompt and seed through both decoders and the pair is indistinguishable at a glance, with marginally softer fine texture in the distilled one. That is one prompt at one resolution, so it is not a quality verdict; it is consistent with keeping the full VAE for saved and refine output, which this PR already does.

Two things to settle before it merges:

- The branch conflicts with `main` in four files (`worker/worker/engine.py`, `worker/worker/manifests.py`, and both test modules), because today's merges moved the engine underneath it. I did not resolve them: rebasing someone else's draft design risks changing what it means.
- The numbers in the description cannot both hold. A 184.9 ms end-to-end median with three calibrated slots is impossible, since slots are `floor(500 / p95)` and p95 is never below the median: 184.9 ms supports two. Three needs p95 at or below 166.67 ms, and the 148.8 ms p95 measured here does reach it, so the claim is probably right and the evidence quoted for it is the wrong statistic. Reporting the p95 fixes it.

leonfullxr:
Superseded by #316, which merged as 5459165. Closing this rather than rebasing it, and leaving its branch alone in case you want the history.

Everything here carried over. I checked rather than assumed before closing: every symbol this draft added is present on main with at least as many references (`preview_decoder`, `_preview_decoder`, `_defer_preview_decoder`, `_detach_preview_decoder`, `PREVIEW_DECODER_RETRY_S`, `_PREVIEW_DECODER_ATTR`, `_PREVIEW_DECODER_RETRY_ATTR`), and every test whose name mentions the preview decoder is there too, with none dropped.

Why it needed reapplying rather than rebasing. This branch was 100 commits behind, with conflicts in `engine.py`, `test_engine.py`, `test_manifests.py`, `manifests.py`, `architecture.md` and `decisions.md`, and it contains a merge commit, so a rebase would have been messier than taking its net diff onto a fresh branch. That part was mechanical.

The part that mattered is that this draft predates the T2I-Adapter conditioning change. `_frame` now splits on `manifest.t2i_adapter`: the adapter branch renders and returns before reaching `_preview_decoder`, which sits in the image-to-image branch below it. Every shipped realtime model declares an adapter, `vega-rt` and `sdxl-turbo` both, so as reapplied this decoder was never used at all. `vega-rt` measured 335.2 ms p95 with it configured and 340.5 with it removed, the same number, while 154 tests passed either way. A clean rebase and a green suite would both have said this was ready.

Three things changed on the way in, beyond wiring the decode into the adapter branch. The decoder detach moved from each eviction caller into `_drop_pipeline`, because every drop path routes through there and a decoder outliving its pipeline pins VRAM for weights already gone. The `wire()` exclusions and the architecture text now list `preview_decoder` and `t2i_adapter` together, since both are worker-side. And this branch's decision entry claimed the distilled decoder moved calibration from one slot to three off a 184.9 ms frame; `floor(500 / 184.9)` is two, and a third would need a frame under 166.7 ms, so every figure in that entry has been checked against `slots_from_frame_ms` and the entry now also carries what the shipped conditioned path measures, 216.5 ms p50 and 219.8 p95 against 331.9 and 336.4.

Two things this work makes load-bearing are filed rather than fixed: #317, because clearing the calibrated slot count on a deferral cannot reach a live worker while `realtime_slots` crosses the wire only in `hello`, and #319, because no model reference in the repository pins a revision and this adds one more unpinned one.

PR #262, opened 2026-08-12
Give each worktree its own dev ports and database

### LINKED ISSUE #261: Two checkouts cannot run the dev stack at once
# Description

Two checkouts cannot run the dev stack at the same time: the ports are fixed and the dev database is shared.

# Motivation

Parallel worktrees are routine here, and the test database already derives its name from the checkout path so suites do not collide. The dev loop did not get the same treatment.

`API_PORT ?= 8000` and `WEB_PORT ?= 5173` exist in the Makefile and `make dev-start` honours them, but `make api` hardcodes `--port 8000` and `ALLOWED_ORIGINS=http://localhost:5173`, and `frontend/vite.config.ts` now sets `strictPort: true`, so a second `make web` fails outright rather than drifting to another port. Working around it by hand is what a session already does.

The dev database `potocolom` is shared by every checkout, so two sessions running migrations or generating share one set of rows.

# Functionality

- `make api` and `make web` honour `API_PORT` and `WEB_PORT` rather than hardcoding them, including in `ALLOWED_ORIGINS` and the vite `server.port`.
- Default those ports from a hash of the checkout path, the way `backend/tests/conftest.py` already derives its database name, so two worktrees differ without anyone setting anything.
- Derive the dev `DATABASE_URL` the same way, and make `make test-db-clean` drop the dev databases too.
- One helper computes the suffix; conftest uses it rather than repeating the hash.

# Details

Keep the variables overridable. A developer who wants :8000 should still get it by asking, and CI sets nothing so it must keep working from the defaults.

`strictPort: true` stays: failing loudly on a taken port is right, and it is only a problem because the port is fixed. This makes the failure impossible rather than removing the check.

Out of scope: the dev containers themselves. PostgreSQL, Redis and MinIO stay one shared instance per machine; this is about which database inside it a checkout uses.

leonfullxr:
Closes #261. Implemented by OpenCode (`deepseek-v4-flash`, max effort) against a design I fixed in the brief; I verified everything below myself because the delegate was cut short before it ran the suite.

Parallel worktrees are routine here and the test database already derives its name from the checkout path. The dev loop did not: `make api` hardcoded `--port 8000` and an `ALLOWED_ORIGINS` of `:5173`, `vite.config.ts` hardcoded 5173 with `strictPort: true`, and every checkout shared the `potocolom` dev database. A session was already working around this by hand on :4200.

**The main worktree keeps 8000, 5173 and the bare database**, so the ports documented in `local-development.md` stay true for the ordinary single checkout. Only a linked worktree derives an offset, which is where parallel sessions live. The bands 8100-8499 and 5700-6099 avoid every port this machine and the dev containers already hold.

`strictPort` stays. Failing loudly on a taken port is correct and was only a problem because the port was fixed; this makes the collision impossible rather than removing the check.

## Verified here

```
main worktree     API_PORT=8000 WEB_PORT=5173 DB_SUFFIX=
linked worktree a API_PORT=8355 WEB_PORT=5955 DB_SUFFIX=_48dffad8
linked worktree b API_PORT=8265 WEB_PORT=5865 DB_SUFFIX=_0eb51875
```

In a linked worktree with the new Makefile, `make api` resolves to `--port 8306`, `ALLOWED_ORIGINS=http://localhost:5906` and `.../potocolom_a34edc1d`, and `make web` passes both ports through. `API_PORT=9999 make api` still wins, so the override survives.

`test-db-clean` matches an exact eight-hex-character suffix, so it drops the per-worktree databases and spares `potocolom`, `potocolom_bench` and `potocolom_sections`. Confirmed against the live list.

`make verify` green: backend 233, worker 90, frontend build and CSP. No non-ASCII.

## Note on the delegate

It was interrupted by a sandbox permission prompt while cleaning up its own test worktree, so it never reached the test suite and left the worktree behind. I removed it and ran the gate myself. Its earlier claim that the main worktree prints canonical values was correct, but the linked-worktree check I repeated independently rather than take on trust: its own run had used main's Makefile, which cannot exercise the change at all.

## Summary by CodeRabbit

- **New Features**
  - Added automatic, checkout-specific API and web development ports for linked worktrees.
  - Added matching database suffixes to isolate linked-worktree development and test databases.
  - Added a streamlined workflow for creating per-checkout development databases.
  - Added environment-variable overrides for development ports and database connections.

- **Bug Fixes**
  - Improved database cleanup while preserving the shared development database.
  - Updated frontend proxying, API startup, and benchmarking to use configured ports.

- **Documentation**
  - Expanded local development guidance for configurable ports, databases, and worktree setup.


leonfullxr:

Issue #280, opened 2026-08-13
Two pytest runs on one checkout corrupt each other through the scratch database


leonfullxr:
# Description

Two pytest runs against the same checkout share one scratch database and corrupt each other's job state. The failures land on the queued-job retry tests and look like application defects: an upload key that was just issued is refused with 403, or a job never reaches the attempt the test waits for.

Reproduced on `main` (0f931ec), with no local changes, by starting the full suite and then the jobs file three seconds later against the same checkout:

    full suite : 1 failed, 237 passed in 198.76s
    jobs file  : 10 failed, 44 passed in 41.44s

Each run passes on its own. The same shape appeared twice while reviewing unrelated work, costing an investigation each time:

    FAILED tests/test_jobs.py::test_retried_attempt_uses_a_new_upload_key_and_rejects_the_old_key
    FAILED tests/test_jobs.py::test_a_retry_does_not_leave_the_earlier_attempt_behind
    FAILED tests/test_jobs.py::test_a_reported_failure_does_not_leave_its_upload_behind
    FAILED tests/test_jobs.py::test_a_late_verdict_does_not_fail_the_attempt_that_replaced_it
    FAILED tests/test_jobs.py::test_a_rejected_upload_is_not_left_in_storage
    FAILED tests/test_jobs.py::test_a_rejected_retry_collects_every_attempt
    FAILED tests/test_jobs.py::test_malformed_completion_is_recoverable_through_the_fleet_socket

    assert client.put(second_path, content=png_bytes()).status_code == 200
    AssertionError: assert 403 == 200

The 403 is the honest answer to a stale key: the other run's rows advanced the attempt the test was holding.

# Motivation

`backend/tests/conftest.py` gives each checkout its own database precisely so concurrent worktrees cannot arbitrate over one, and its docstring explains why. Per process is the case it does not cover, and that case is ordinary: an editor test runner while a terminal run is going, an agent driving probes while someone runs the suite, or two terminals. The symptom points at the retry logic rather than at the harness, which is what makes it expensive.

# Functionality

Make a second concurrent run either safe or refused, rather than silently destructive:

- Isolate per process rather than per checkout, for example by extending the database name with something unique to the run, and dropping it at exit.
- Or take an advisory lock for the suite's lifetime and fail fast with a message naming the other run, which is honest and much cheaper than isolation if concurrent runs are not worth supporting.
- Either way, the tests that break should not be the ones that look like application bugs.

# Details

[...]

PR #287, opened 2026-08-13
Keep a job tracked until its terminal state is durable

### LINKED ISSUE #248: Terminal job paths de-track before the database transition is durable
# Description

Terminal job paths remove the in-memory entry before the database transition is durable, so a transient database failure leaves the row `running` with nothing tracking it.

# Motivation

#204 moved worker-supplied field conversion before `inflight.pop`, which fixed the case where a malformed value stranded the job. The ordering problem underneath it is unchanged: the pop, the watchdog removal and `release_job_slot` all happen before the commit that writes the terminal state.

A lock timeout, a flush error or a commit failure after that point leaves the row `running` forever. `on_worker_lost` cannot requeue an entry that is gone, and `sweep_stalled_jobs` only walks `inflight`, so neither recovery path can see it. This affects successful completion, the invalid-output path and `job_failed` alike.

It needs a real database failure to trigger, which is why it survives normal testing, but a lock timeout under load is exactly when it matters.

# Functionality

- Keep the in-flight entry until the terminal transaction commits.
- Compare and pop afterwards, so a requeue that happened while the transaction was open is not undone.
- On database failure, leave the entry in place and let the existing recovery paths retry rather than dropping the job on the floor.

# Details

The slot accounting has one writer by design, so moving `release_job_slot` after the commit needs care: the slot must not stay held if the process dies between the commit and the release. Releasing after a successful commit and accepting a lost slot on a hard crash is probably right, since worker reconnection rebuilds that state anyway, but that is worth stating explicitly when this is implemented.

Found by adversarial review of PR #243, which fixed the conversion ordering but not the durability ordering.

leonfullxr:
Closes #248. Implemented by OpenCode (`deepseek-v4-flash`, max effort) to a design I fixed in the brief; gated and mutation-checked by me.

Every terminal path popped the in-flight entry, the watchdog stamp and the worker slot **before** the transaction that writes the terminal state:

```
entry = inflight.pop(job_id, None)
...
release_job_slot(worker)
... then the transaction that writes succeeded or failed
```

A lock timeout, a flush error or a commit failure after that point left the row `running` with nothing tracking it. `on_worker_lost` cannot requeue an entry that is gone and `sweep_stalled_jobs` only walks `inflight`, so neither recovery path could ever see the job again. It takes a real database failure to reach, which is why it survived normal testing, and a lock timeout under load is exactly when it matters.

## What changed

The entry lives until the commit. Inside the transaction the row is checked for still belonging to this attempt (gone, already terminal, or an attempt past this one means a requeue or another verdict won), and the verdict is refused rather than overwriting it. After a successful commit the entry is cleared only if it is still this one, so a requeue that landed mid-transaction keeps both its entry and its slot. A transaction that raises leaves everything in place for the existing recovery paths.

`mark_failed` reports whether it committed, so callers can tell a durable failure from one that did not happen.

The slot is released after the commit, so a process dying in between leaks one slot until that worker reconnects. Accepted trade, and the comment says so, because the obvious tidy-up is to move it back.

## One deviation from my brief, surfaced by the delegate

I specified `job.attempt == current.attempt`. That breaks three existing tests which install a fake attempt-2 entry while the row still says attempt 1. The implemented rule is `job.attempt > current.attempt` refuses. The row attempt only ever grows, so it is the same guard against a requeue winning the race, and it does not depend on the tests fiction of an entry ahead of its row. I checked it rather than taking it on trust.

## Guards proved by breaking them

| Guard | Mutation | Result |
|---|---|---|
| Clear state only when the entry is still ours | back to unconditional pop plus release | test fails |
| Attempt check inside the transaction | reduced to `job is None` | test fails |
| Keep the entry when the transaction raises | pop on the exception path | tests fail |

`make verify` green: 260 backend, 93 worker, frontend build and CSP.

**Note on the suite:** an interleaved second `pytest` run produced 11 failures that are not real. Two concurrent runs on one checkout corrupt each other through the shared scratch database, which is issue #280; a single clean run passes 260.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery of interrupted jobs, allowing failed recovery attempts to be retried safely.
  * Prevented older job attempts from overwriting newer progress or results.
  * Ensured worker capacity is released reliably and only once after job completion.
  * Improved handling of database, storage, and output-link failures so completed jobs retain their final status and usage records.
  * Reduced the risk of jobs becoming permanently stalled during concurrent retries or failures.
  * Improved accuracy of dimensions reported for completed assets.


leonfullxr:
Full `make verify` on a quiet machine after merging current main: **307 backend, 134 worker, frontend build and CSP, green.** CI is green on all five checks.

Review loop closed at round 2 with nothing medium or higher, so from my side this is ready.

leonfullxr:
Third opinion from `opencode-go/kimi-k3` (read-only, after Codex twice and CodeRabbit): **no high or medium findings**, four lows, one refactor. Three taken in b087fb1.

It answered the interleaving questions by reading `realtime.py` rather than reasoning in the abstract, which is what made the answers worth having:

- **Two verdicts for one job cannot start together.** The fleet loop awaits `on_worker_message` inline, so a second message is not read until the first verdict has committed, cleared and released. Across a reconnect the `Worker` object differs, so the identity check rejects the stale one, and the row lock plus the attempt check refuses it even if that passed.
- **A verdict cannot race `on_worker_lost` from the same connection.** The socket death only surfaces at the next `receive`, and the disconnect cleanup runs in the `finally` strictly after the handler returns. Either the verdict commits and clears, or it fails and the surviving entry is what `on_worker_lost` hands to `lost_jobs`.
- **Against the stall sweeper**, both the sweeper pop-and-release and the post-commit clear-and-release are await-free, so exactly one wins the entry and exactly one release happens; the loser is refused on the database side by the attempt check.

It also corrected my own comment: the accepted "slot leak between commit and release" is narrower than I wrote, since there is no await between them, so only a hard kill separates the two, and a kill rebuilds slot state anyway.

## Taken

| Finding | Fix |
|---|---|
| Low: `last_progress_at` never cleared when dispatch fails on a missing source asset | The stamp is popped with the entry. Nothing collected those: the sweeper walks `inflight`, so each one stayed for the process life |
| Low: a comment in `_worker_int` justified its int4 clamp with "the commit below is not inside a try", which this branch made false | Rewritten to the reason that now applies |
| Refactor: the clear-and-release block appears three times, guard comment included | One `clear_if_current` helper, invariant stated once. Three copies of an exactly-once rule will drift |

## Not taken, deliberately

- The duplicate `db.session_factory is None` check is harmless and only worth touching while editing that function anyway.
- The two uncovered branches it names (the failure-path refusal and the invalid-image commit failure) are structurally identical to covered paths, so a test there guards against a future regression rather than present behaviour. Worth adding when that function is next touched.

`make verify` green after the change: 308 backend, 134 worker.

One process note, since it bit twice today: a suite run that takes 14 minutes instead of 3 and fails 9 tests here is the CI runner working on the same machine, not a regression. A clean rerun passed all 75.

leonfullxr:
The failing `backend` check does not reproduce locally, which makes it an isolation problem rather than a regression in this branch.

`test_retried_attempt_uses_a_new_upload_key_and_rejects_the_old_key` passes on its own against this branch, and the full backend suite passes too: 308 tests, no failures, run with `PYTHONPATH=$PWD ./.venv/bin/python -m pytest -q` on the branch head with the merge of main already in.

CI failed two assertions in that area, both a fresh dispatch's upload token answering 403 where 200 was expected, at `tests/test_jobs.py:782` and `:819`. A token that is valid locally and rejected in CI points at state shared between tests or between runs rather than at the token logic itself.

Worth noting #292 is open for exactly this class of problem ("Give each test run its own database, not each checkout"). If that is the cause, this branch is not the place to fix it, and re-running the check after #292 lands would confirm it either way.

I have not touched this branch. Flagging it because a red check on a durability PR reads as a real defect and this one appears not to be.

leonfullxr:
The `backend` check that went red here was `403 == 200` in the retry tests, which is issue #280 rather than anything on this branch: the tests leave `JOB_STALL_SECONDS=0.05` armed while they assert, so on a loaded runner the sweeper fires again and fails the row past its one retry. This branch does not touch those tests.

I fixed it on #292, which is the isolation PR, by disarming the sweeper once a test has the retry it was waiting for. Eight concurrent runs of the worst offenders pass there. Once that merges this class stops costing anyone a red check; in the meantime a rerun here is green.

PR #289, opened 2026-08-14
Finish every delete batch before reporting what failed

### LINKED ISSUE #256: A partial DeleteObjects failure abandons the remaining version batches
# Description

`S3Storage.delete` raises on the first `DeleteObjects` batch that reports an error, so later batches for the same key are never attempted.

# Motivation

The purge pages `list_object_versions` and deletes in batches of a thousand, which is the `DeleteObjects` cap. A key with more than a thousand versions therefore takes several batches. If the first batch returns a single transient `InternalError`, the successful deletions in that batch stand, the raise aborts the rest, and the remaining versions stay. The caller logs a warning and moves to the next key, so nothing returns to them.

Raising is right: the earlier behaviour discarded the response entirely and reported a reclaim that had not happened. Aborting the traversal is the part that is wrong.

# Functionality

- Attempt every page and every batch, accumulating the failure count and the first error.
- Raise once at the end when anything failed, keeping the message shape the caller already logs.

# Details

Low severity: it needs both a key with over a thousand versions and a partial batch failure. Neither is likely in the shipped profile, where a key accumulates one version per attempt and jobs retry once. It matters more if a future change makes versions accumulate faster, and the fix is small enough that it is not worth waiting for that.

Related: #254 covers retrying a failed cleanup at all, which is the larger gap. This issue is only about not giving up early inside one call.

Found by adversarial review of PR #243.

leonfullxr:
Closes #256.

`S3Storage.delete` pages the version list and deletes in batches of a thousand, the `DeleteObjects` cap, and raised on the first batch that reported an error. The successful deletions in that batch stood, every later batch and every later page were abandoned, and the caller only logs a warning, so a key past a thousand versions could keep most of its versions after one transient `InternalError` with nothing coming back to anyone.

Raising is still right: discarding the response reports a reclaim that did not happen. It now happens once at the end, counting failures from every batch and naming the first code seen, so the logged message describes the whole shortfall rather than one batch of it.

The fake S3 client gained per-call responses, which is what lets a test fail one batch and watch the batches behind it still run.

## Guards proved by breaking them

| Guard | Mutation | Result |
|---|---|---|
| Traversal continues past a failed batch | raise restored inside the loop | both new tests fail |
| Failures accumulate across batches | `failed +=` back to `failed =` | the counting test fails |

`ruff`, `mypy` and `backend/tests/test_storage.py` (26 tests) are green.

**Not yet run: the full `make verify`.** Another process on this machine is using the GPU and the CPU is contended, so I have deliberately kept to the sub-second storage suite for now and will run the full gate before this is merged. CI on this PR runs it independently in the meantime.

Implemented directly rather than delegated: the OpenCode run died two tool calls in, which is its third failure of this kind today, and the change is about fifteen lines.

## Summary by CodeRabbit

* **Bug Fixes**
  * S3 version cleanup now continues processing remaining batches and pages when individual deletions fail.
  * Cleanup reports one aggregated error with the total failure count and the first error code encountered.
  * Successfully deleted objects are no longer retained after partial failures.

* **Tests**
  * Added coverage for multi-batch deletion, pagination, partial failures, and aggregated error reporting.


leonfullxr:

leonfullxr:
Third opinion from `opencode-go/kimi-k3` (read-only, after Codex): **no blockers**, three lows and a nit. Three taken.

The useful half of this review was about what the tests do not pin, which is the failure mode I care about most here:

- **Nothing asserted the listing uses a prefix.** A refactor to `paginate(Bucket=...)` with no `Prefix` would have kept all four delete tests green (the fake filters by whatever prefix it is given, and `startswith("")` is true for everything), while turning every delete into a full scan of the bucket version history. Pinned; removing the prefix now fails.
- **Nothing asserted `Quiet=True`.** Losing it changes no test and gives production a thousand-entry success list per batch. Pinned in the fake so every test enforces it; five fail without it.
- **The raised message dropped the error `Message`.** The caller only logs this line, so it is the entire artefact of a failed reclaim: the code is what an operator greps for, the message says which permission is missing. Both are in it now.

Not taken, with the reason on the record: the page walk runs to the end of the prefix rather than breaking at the first page carrying another key. An early break is available (versions of one key are contiguous, and this key sorts before anything extending it), and it buys nothing for this repository, where keys diverge before the extension so a purge sees one page. kimi called skipping it defensible and asked for a comment; there is one.

It also confirmed the things I would otherwise have had to take on trust: the pagination cannot loop or grow unbounded because the marker advances past deleted versions and `doomed` is rebuilt per page; a `ClientError` still escapes after SDK retries; the retry contract is idempotent because `delete` is whole-key; and the `.bak` fixture fix is real, since `"u/j-attempt-10.png".startswith("u/j-attempt-1.png")` is False and the old collision was fiction.

On the raise-at-the-end contract it agreed with the current shape: both callers log and continue, `LocalStorage.delete` also raises, and returning a count instead would ripple through the protocol for no benefit.

`make verify` green: 303 backend, 134 worker.

PR #291, opened 2026-08-14
Read an unregistered worker as current, not as ancient

### LINKED ISSUE #282: Worker.protocol_version is optional, so an unset version fails open
# Description

`Worker.protocol_version` in `backend/app/realtime.py` is `int | None` with a default of `None`, and the dispatch-token gate in `backend/app/jobs.py` reads it as `worker.protocol_version is not None and worker.protocol_version >= 3`. A `Worker` built without a version is therefore treated as a protocol 2 worker and its tokenless control messages are believed.

# Motivation

Registration always sets the version from the validated hello, so nothing in production reaches this today; the exposure is a future construction path, and the unit tests that build a `Worker` directly. A security gate that reads an unset field as the permissive value is the wrong default whether or not it is currently reachable.

# Functionality

Make `protocol_version` a required constructor argument and give every test worker an explicit version, or default it to `PROTOCOL_VERSION` so an unset value fails closed rather than open.

# Details

`backend/app/realtime.py`, the `Worker` dataclass and the `fleet` handler; the consumer is `on_worker_message` in `backend/app/jobs.py`. Several tests in `backend/tests/test_jobs.py` and `backend/tests/test_realtime.py` construct `Worker` directly and would need the argument.

Found by review of PR #277 (codex gpt-5.6-sol, xhigh), round 4, which was otherwise clean.

leonfullxr:
Closes #282.

`Worker.protocol_version` defaulted to `None`, and two gates read it: the dispatch-token check in `jobs.py` requires the token from a worker at 3 or newer, and `update_params` in `realtime.py` refuses an update an N-1 worker would silently drop. `None` meant "predates every version" for both, so the security gate handed an unset version exactly the leniency that exists only for an older worker.

Registration always sets the field, so nothing in production reached this. A security gate that reads an unset field as the permissive value is the wrong default regardless. It defaults to the current protocol now and both consumers compare it directly, which also drops the `(x or 0)` at the `update_session` call site.

The tests that build a `Worker` without a registration were leaning on that leniency to send job messages with no token, so they were exercising the acceptance rather than the numeric conversion they were written for. They carry a token now, and a new test pins the default: with `protocol_version` forced to `0`, it fails.

## Scope note

Full `make verify` deliberately not run: another process on this machine holds the GPU and the CPU is contended. What I ran is `tests/test_realtime.py` (113 tests), the affected `tests/test_jobs.py` selections, `ruff` and `mypy`, all green, plus the mutation check above. CI runs the full gate here, and I will run it locally before this merges.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved worker protocol compatibility checks for dispatch tokens and live parameter updates.
  - Unregistered workers now use the current protocol version by default.
  - Tokenless progress messages from workers requiring tokens are rejected consistently.

- **Tests**
  - Updated worker-message handling tests to include dispatch tokens.
  - Added coverage for default protocol-version behavior.

### DIFF
--- a/backend/app/jobs.py
+++ b/backend/app/jobs.py
@@ -1227,8 +1227,7 @@ async def on_worker_message(worker: realtime.Worker, control: dict) -> None:
     # believed (docs/connection-handling.md), which is the compatibility
     # floor; that acceptance disappears when the floor moves to 3.
     presented = control.get("dispatch_token")
-    if (presented is None and worker.protocol_version is not None
-            and worker.protocol_version >= 3):
+    if presented is None and worker.protocol_version >= 3:
         # A current worker omitting the token is as stale as one presenting a
         # wrong token, and gets the same warning.
         logger.warning("worker %s sent a stale dispatch token for job %s; ignored",
--- a/backend/app/realtime.py
+++ b/backend/app/realtime.py
@@ -305,11 +305,15 @@ class Worker:
     realtime_slots: int
     device: str | None = None
     memory_mode: str | None = None
-    # Advertised in hello. The update_params handler compares it against
-    # UPDATE_SESSION_PROTOCOL_VERSION to refuse an update an N-1 worker would
-    # silently drop. None for a Worker built without a registration, which is
-    # read as predating every version.
-    protocol_version: int | None = None
+    # Advertised in hello, and read by two gates: the update_params handler
+    # compares it against UPDATE_SESSION_PROTOCOL_VERSION to refuse an update
+    # an N-1 worker would silently drop, and the dispatch-token check in
+    # jobs.py requires the token from a worker at 3 or newer. A registration
+    # always sets it; the default covers a Worker built without one, and it is
+    # the current version rather than None so that gate fails closed rather
+    # than granting the leniency that exists only for an older worker
+    # (issue #282).
+    protocol_version: int = PROTOCOL_VERSION
     slots_in_use: int = 0
     jobs_in_flight: int = 0  # queued jobs; capped at JOB_DISPATCH_DEPTH in jobs.py
     last_seen: float = field(default_factory=time.monotonic)
@@ -822,10 +826,8 @@ async def realtime(ws: WebSocket) -> None:
                                     "message": invalid,
                                 }))
                                 continue
-                        # None means no registration recorded a version, which
-                        # is read as predating every one of them.
                         if (session.worker is not None
-                                and (session.worker.protocol_version or 0)
+                                and session.worker.protocol_version
                                 < UPDATE_SESSION_PROTOCOL_VERSION):
                             # The assigned worker predates update_session (an
                             # N-1 worker is deliberately welcome) and would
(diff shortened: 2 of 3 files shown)


leonfullxr:
Review round 1 (codex `gpt-5.6-sol`, xhigh, read-only): **ready to merge, no findings.**

Worth recording what it checked, since the interesting question here was whether fixing one gate broke the other:

- The current-version default is right for **both** consumers. A missing version now means an in-process worker that is current by construction, not an unknown network peer, so the token gate holds it to the token and the `update_session` gate sends it the update. That would be the wrong reading for a real peer of unknown version, and that state cannot occur: `hello` requires the field and the only shipped `Worker(...)` construction passes it.
- No shipped code builds a `Worker` without a version. Every other construction is a test.
- The tokens added to `test_non_finite_progress_is_ignored` and `test_job_done_with_unusable_numbers_leaves_the_job_recoverable` do not short-circuit what those tests were written for: both reach their number-conversion paths, and the surviving `0.5` assertion proves the progress test does not stop at the token check.
- The new default test is not a tautology: it pins the field default and its security effect together.

One correction it made to me, applied in the follow-up commit: my docstring said both gates refuse an unregistered worker. Only the token gate does. The update gate reads the same field to decide whether a worker is new enough to receive an update, so the same default makes it send rather than withhold, which is the correct answer for a worker that is current by construction.

Still outstanding: full `make verify` on a quiet machine. What ran here is `tests/test_realtime.py` (113), the affected `test_jobs.py` selections, `ruff`, `mypy`, and the mutation check that forcing the default to `0` fails the new test.

leonfullxr:

leonfullxr:
Third opinion from `opencode-go/kimi-k3` (read-only, after Codex): **clean, no findings.** "The default fails closed for the token gate and open for the update gate, which is the right direction for each."

What it checked that is worth having on the record, since it read the validation rather than assuming it:

- **Nothing malformed reaches either gate.** Every check runs in `fleet` before the worker is registered: a missing key closes with `CLOSE_PROTOCOL_VIOLATION`, a string, float, list or null fails the `isinstance` check, a negative fails the floor, and an integer past CPython's digit limit is rejected by `parse_control` as a protocol error. A `Worker` object that fails validation is discarded before registration, so the half-built one never escapes.
- **One cosmetic oddity, not worth fixing:** `true` passes `isinstance(version, int)` and is then rejected by the floor, so it closes with `CLOSE_UNSUPPORTED_VERSION` rather than `CLOSE_PROTOCOL_VIOLATION`. Right outcome, slightly wrong label.
- **A far-future version registers and is treated as current or newer**, which is deliberate: both gates are lower-bound comparisons and a parametrized test already registers `PROTOCOL_VERSION + 1` and asserts updates flow.
- **N-1 traced end to end from the worker client**, not assumed: a protocol 2 worker registers (the floor is inclusive), ignores the extra `dispatch_token` in its dispatch, and its token-less reports pass the gate. `update_params` is refused on its behalf with a browser-facing error rather than silently dropped, which is the designed behaviour.
- **Repo-wide, only two shipped sites read the field**, plus the validation and the client that writes it into `hello`.

It surfaced one doc/code gap that predates this PR: `docs/blueprint.md` lists `protocol_version` in the `worker:{worker_id}` Redis hash, and nothing persists it, because that Redis layout is not implemented. I am correcting that in the documentation branch (#297) rather than here.

`tests/test_realtime.py` 113 passing under its own run.

Issue #279, opened 2026-08-13
Tests inherit each other's running jobs


leonfullxr:
# Description

Backend tests that leave a job running make the next test inherit it. When a test socket closes, the disconnect handler requeues whatever that worker held, and the next test to open a fleet socket is handed those dispatches before its own.

# Motivation

Two tests added in PR #277 were flaky for exactly this reason, and the fix there was a helper, `dispatch_for`, that reads until it sees its own job id. That makes the new tests deterministic and leaves the underlying leak in place: the skipped dispatches still occupy the worker dispatch depth, are requeued again when the socket closes, and can in principle fill both slots so the requested job never arrives.

# Functionality

Either make every test terminalize the jobs it creates, or give the suite a fixture that clears queued and running rows plus the in-process dispatch state between tests, and then drop the skipping helper.

# Details

`backend/tests/test_jobs.py`, the `dispatch_for` helper and the tests that use it. The state involved is the `jobs` table plus `jobs.inflight`, `jobs.live_progress` and `jobs.last_progress_at`.

Found by review of PR #277 (codex gpt-5.6-sol, xhigh).

leonfullxr:
#306 landed and closes most of this: an autouse fixture clears the job rows, the in-process dispatch state, the queue, the worker registry and the settings cache after every db test. Two follow-ups were needed and are also in: #307 (delete assets before jobs, since `DELETE` does not absorb what `TRUNCATE CASCADE` did) and #308 (force the retry rather than waiting for the sweeper).

Measured effect: the backend suite went from about 145 seconds to about 45, because tests were waiting on each other's leftovers, and the `403 == 200` cluster that cost CI runs on three unrelated branches is gone.

**Left open deliberately, which is why this issue is not closed:** `dispatch_for` stays. I removed it, measured, and it fails about one run in four, always the same cluster. The remaining cause is a dispatch already in flight when the fixture runs: the loop picked the row before the delete and delivers it to whichever worker is registered when it wakes, which is the next test's. Its docstring now says that rather than describing the leak that is gone.

Closing that means either draining the dispatch loop at teardown or having `dispatch` re-check that its worker is still the one the job was picked for. Both are real changes to production code paths for a test-only symptom, which is why I stopped here rather than reaching for them.

leonfullxr:
Still open for the remainder. `dispatch_for` stays. Draining production dispatch for a test-only race was already rejected. This campaign does not change that.

PR #292, opened 2026-08-14
Give each test run its own database, not each checkout

### LINKED ISSUE #280: Two pytest runs on one checkout corrupt each other through the scratch database

leonfullxr:
Closes #280.

The scratch database carried a hash of the checkout path, which keeps worktrees off each other's migrations. Two runs of one checkout still shared it, and that case is just as ordinary: an editor test runner beside a terminal run, or an agent probing while the suite goes.

They shared job and attempt rows, so the retry tests failed with a 403 on an upload key that had just been issued, or waited for an attempt the other run had already advanced. Every one of those failures reads as an application defect. It cost an investigation twice during review, and twice more here yesterday: two double-digit failure runs on the #248 branch that were entirely this, and it sent the delegate chasing ghosts as well.

The name now carries the process id, and the run drops its own database in `pytest_sessionfinish`. Cost is a migration chain per run rather than per checkout, a second or two.

## Details worth knowing

- The drop is a session hook rather than `atexit`: asyncpg needs an event loop and a thread pool, and scheduling one at interpreter shutdown raises `cannot schedule new futures after interpreter shutdown`. I hit that first and moved it.
- An exported `DATABASE_URL` is left alone and dropped by nothing. That is the developer's database, not this file's, and the flag recording which one it is has to be read before `setdefault` runs.
- A hard kill still leaves a database behind. `make test-db-clean` already matches these with its `LIKE potocolom\_test%`, so nothing there needs changing on this branch. **Interaction with #262:** that PR tightens the pattern to `^potocolom_test_[0-9a-f]{8}$`, which would stop matching the new names. I have pushed the widening (`(_[0-9]+)?`) onto that branch so whichever lands second is still correct.

## Verified

Two concurrent runs of `tests/test_estimates.py` both pass, and a live check during a run shows two distinct `potocolom_test_<hash>_<pid>` databases. `tests/test_db.py`, `tests/test_migrations.py` and `tests/test_estimates.py` pass, and the per-run database is gone afterwards. `ruff` clean.

**Full `make verify` deliberately not run:** the GPU box is busy and the CPU is contended, and this change is precisely about what concurrent heavy runs do to each other, so it deserves a clean machine. CI covers it; I will run the full suite here before it merges.

## Summary by CodeRabbit

- **Tests**
  - Improved test database isolation with unique names for parallel or repeated test runs.
  - Added clearer reporting for connectivity, PostgreSQL, and migration issues.
  - Supplied databases are preserved without truncating existing data.
  - Automatic cleanup removes only databases created during the test session.
  - Cleanup errors no longer obscure test results.

- **Documentation**
  - Clarified local development setup requirements for using a separate test database.
  - Expanded database configuration guidance for development and test environments.


leonfullxr:
Review round 1 (codex `gpt-5.6-sol`, xhigh, read-only): one high, one medium, one low. All three answered in b3b5f9e except the xdist half of the low, which is deliberately left alone.

**High, and it predates this branch.** `_prepare_database` drops and recreates a database stamped at a revision the tree cannot resolve. That is right for the name this file invents and wrong for one a developer or CI exported: the `_OURS` guard I added only covered the session-end drop, so a supplied database was still destroyed during setup. It now refuses with a message naming the database, and the URL is captured once at import so a later environment change cannot redirect either path.

Proved both directions rather than reasoning about them:

```
supplied database, healthy      -> 1 passed, database still there afterwards
supplied database, stamped 9999 -> RuntimeError: ... not mine to drop; point DATABASE_URL
                                   elsewhere or migrate it yourself
                                   alembic_version still 9999, database not rebuilt
```

**Medium: a false green.** Every `asyncpg.PostgresError` became `DATABASE_AVAILABLE = False`, so a role without `CREATE DATABASE`, a failed truncate, or two runs racing a name skipped all 74 db tests under the message "PostgreSQL unreachable". In CI, which supplies its own URL, that is a green job with no database coverage in it. Only a connection failure against this file's own name still skips; anything else raises, and an unreachable supplied URL raises too:

```
DATABASE_URL=...:5999/nope -> RuntimeError: DATABASE_URL was supplied but unusable: [Errno 111]
```

**Low, half taken.** The name gains a random tail, because a pid is recycled and two containers sharing this checkout can hold the same one, which would let one run force-drop another's database. The pid stays since it names the run for anyone watching `pg_database`.

The xdist half is not taken: under `pytest-xdist` the controller exports `DATABASE_URL` and every worker inherits one database. Nothing here runs xdist, so a fix now would be guessing at its shape, and the commit message says so rather than leaving it silent.

Also confirmed by the round, so it does not need re-deriving: `pytest_sessionfinish` covers normal exit, `-x`, collection errors, Ctrl-C after session start and `--collect-only`; `TestClient` shutdown calls `db.dispose()` before the drop and `NullPool` holds no idle connections; nothing in the suite depends on state surviving between runs; and this does not make #279 harder to fix, only easier to diagnose.

Still not run here: the full `make verify`. The machine is contended and this change is about what concurrent heavy runs do to each other, so it gets a clean machine before merge.

leonfullxr:
Full `make verify` now run on a quiet machine: **301 backend, 134 worker, frontend build and CSP, green.** Branch merged with current main first.

One thing happened while doing it that is worth recording here, because it is evidence for this PR rather than against it. The first verify run on another branch produced eleven failures, all in the retry tests, exactly the list in issue #280. Nothing else was running in my terminal. The cause was the self-hosted CI runner coming back online and executing jobs **on this same machine** at the same time.

So the collision this PR fixes is not only "two terminals" or "an editor runner beside a suite". On a machine that also hosts the runner, ordinary CI counts as the second run. That makes the per-run naming worth more than the issue estimated, and it is a second independent reproduction of the symptom.

A single clean run afterwards passed 302.

leonfullxr:
Third opinion from `opencode-go/kimi-k3` (read-only, after Codex and CodeRabbit): **no high findings**, two medium, five low. Six taken in c6a59db.

The two mediums were both mine, and one is the kind worth stating plainly:

**My own new test used a fixed database name.** `potocolom_test_supplied_probe`, a constant, inside the branch whose entire purpose is that two runs of one checkout must not share a name. Two runs of that test would drop each other's probe. It is named per run now like everything else.

**The comment promised more than the code delivers.** "Leaves its contents alone" is false: a supplied database is never dropped, rebuilt or emptied, and it is still migrated to head, has the local user row inserted or promoted, has leftover running jobs requeued, and has test rows written into it, because that is what starting the application does. `dev.yml` actively tells developers to export `DATABASE_URL` when 5432 is taken, and nothing told them to use a throwaway name, so someone would have believed the old wording and lost dev state anyway. The comment now states the real contract, and a supplied database that already holds jobs gets a loud warning before the run touches it.

## The rest

| Finding | Action |
|---|---|
| Low: a missing supplied database is created, contradicting "never created"; CI depends on that create | Wording fixed rather than the behaviour, since CI needs it |
| Low: per-run naming made the owned-side rebuild and truncate unreachable, while the docstring still justified them | Both deleted. A generated name never exists beforehand, so those paths could only ever have hit a supplied database, where they are refused |
| Low: the test pinned the truncate guard but not the drop guard | Extended: a supplied database stamped `9999` must raise, and the test fails if that guard is removed |
| Low: an owned database that is reachable but refusing killed the whole collection | Loud only for a supplied URL. For this file's own name the developer keeps the non-db tests, and the skip carries the real reason instead of claiming PostgreSQL is down |
| Low: the documented port-override advice silently opts out of run isolation | One sentence in `dev.yml` and `local-development.md`: give the exported URL a database name of its own |

It agreed that deferring xdist is right, with a caveat worth recording: the day someone adds `-n auto`, workers silently share one supplied database and #280 returns in worker form with nothing pointing at the harness.

`make verify` green after the change: 302 backend, 134 worker. One run before that showed the familiar `403 == 200` failures with the CI runner busy on this machine, which is the very symptom this PR is about.

leonfullxr:
Added the load half of #280, which the issue text flagged and I had left: "the failures are load-sensitive as well as concurrency-sensitive ... whichever fix is chosen, that deadline is worth revisiting."

The retry tests set `JOB_STALL_SECONDS=0.05` to force one requeue, then keep asserting against the attempt it produced **with the sweeper still armed**. On a loaded machine it fires again between reading the second dispatch and uploading to it, fails the row past its one retry, and the assertion sees `403 == 200` on an upload key that was just issued. Same symptom as the shared database, different cause.

This is not hypothetical: it failed CI on #287 while this branch was under review, on a runner sharing a machine with local verify runs. The stall now returns to 600 seconds the moment the retry is observed, in the six places that wait for one.

Evidence: eight concurrent runs of the four worst offenders, all passing. Before this branch, a shared database plus an armed sweeper is what produced the double-digit failures.

`make verify`: 306 backend, 135 worker.

PR #293, opened 2026-08-14
Move a slider by one parameter step per arrow key

### LINKED ISSUE #250: Slider arrow keys move one percent of the track, not one parameter step
# Description

`ParamSliderField` drives its underlying slider on a fixed 0 to 100 track, so one arrow key press moves one percent of a parameter's range regardless of how wide that range is. For a narrow integer parameter the result is that the keyboard barely works: `steps` on the realtime models spans 2 to 8, so a single press moves 0.06 of a step and about seventeen presses are needed to advance from 2 to 3.

Found while driving the realtime canvas sliders in a browser: three arrow presses on Steps left the value at 2 and the parameter sent to the worker unchanged.

# Motivation

The fixed track exists for a good reason, recorded in the component: "Slider track is always 0-100 so bar length stays constant across models." That keeps the bar from changing length when the selected model changes the range, which is worth keeping.

The cost is that keyboard users cannot practically set a narrow integer parameter. Dragging with a mouse is unaffected, and so is reading the value, so this is an accessibility gap rather than a broken feature.

It affects every panel that uses the component, including the generate panel's steps control, not only the realtime canvas.

# Functionality

- An arrow key press should move the parameter by one of its own steps, not by one percent of the track.
- Home and End should still reach the minimum and maximum.
- The bar length must stay constant across models, which is what the 0 to 100 track was for.
- Dragging behaviour should not change.

# Details

The conversion helpers already know each parameter's real step: `ParamRange` carries `step` and `integer`, and `normToValue` rounds to them. The gap is only that the slider's own key handling operates in track units with no knowledge of that step.

One option is to keep the 0 to 100 track and give the slider a step of `100 / ((max - min) / step)` so a press lands exactly one parameter step away. Another is to drive the slider in parameter units and normalise for display. The first is the smaller change.

Worth checking the same question for `guidance` and `strength`, where a one percent move is small but not useless, so the fix should not make those coarser than they are now.

leonfullxr:
Closes #250.

`ParamSliderField` drove its slider on a fixed 0 to 100 track, so an arrow key moved one percent of the range whatever that range was. On the realtime models `steps` spans 2 to 8, so a press moved 0.06 of a step and about seventeen were needed to get from 2 to 3.

The track now carries one notch per parameter step (`trackSteps(spec)`), so a press moves the parameter by one of its own steps, Home and End still reach the ends, and the bar spans the same width whatever the model declares, which is what the fixed track was protecting. Dragging is unchanged.

## On the "do not make guidance coarser" note in the issue

It does not. Guidance moves 0.5 per press instead of 0.19, and 0.19 was never reachable: `normToValue` rounds to the parameter step, so a press that moved less than one step either did nothing or landed on the value the previous press already produced. The reachable set is identical; only the number of presses changed. A test walks the whole `steps` range one notch at a time and asserts `[2, 3, 4, 5, 6, 7, 8]`.

The `edit_image` placeholder keeps the default track: its slider is disabled and has no parameter behind it.

## Verified

`npm test` 58 passing, `npm run lint` and `npm run check` clean. The five new tests were proved by flattening `trackSteps` back to a constant 100, which drops the suite to 53 passing and 5 failing.

Frontend only, so no backend suite is involved; the GPU box being busy does not affect this one. The tests are `node --test` in the house style rather than vitest, which is only present transitively.

## Summary by CodeRabbit

* **Improvements**
  * Parameter sliders now automatically adjust their available steps based on each setting’s range.
  * Generation and real-time controls provide more appropriate precision for steps, guidance, and image-to-image strength.
  * Slider behavior remains stable for fractional, limited, or invalid ranges, with sensible minimum and maximum limits.

* **Accessibility**
  * Screen readers now receive clearer slider value announcements.

* **Tests**
  * Added coverage for slider step calculations and value conversion across supported parameter ranges.


leonfullxr:
Review round 1 (codex `gpt-5.6-sol`, xhigh, read-only): **ready to merge**, one low, which is fixed in the follow-up commit.

**The low, and it is a real one.** `trackSteps` had no upper bound, and the slider (Bits UI 2.18.1) builds an in-memory array with one entry per notch. A manifest declares the range and the step, so a wide range against a tiny step could hand it an arbitrarily large array and freeze the page. Capped at 100, which is exactly the fixed track this replaced: at the cap the keyboard is no worse than before this change existed, and no shipped model comes near it (the widest is guidance at 38). A test pins the cap and fails without it.

**What the round confirmed**, so it does not need re-deriving:

- The bar width really is unchanged: the slider and its track are still `w-full`, and `max` only changes the numeric domain. No caller assumed a 0-100 domain; they all exchange a 0-to-1 normalized value.
- The "this does not coarsen fractional parameters" claim holds for the shipped ranges. Guidance 1-20 step 0.5: 38 notches, the same 39 reachable values. Strength 0-1 step 0.1: 10 notches, the same 11 values. Shipped strength 0.05-1 keeps its old reachable set too, and its inability to select 0.05 already existed in `normToValue`.
- The claim is **not universal**, which is worth stating plainly: when a range is not a whole multiple of its step, the rounded notch count can make a final move shorter, duplicated, or skip a value the old 100-point track could reach. No current manifest has a harmful case, and the honest scope of the claim is "for the ranges we ship".
- The size picker is a toggle group rather than this component, so it is unaffected, and the disabled `edit_image` placeholder correctly keeps the default track.
- The tests are genuine rather than replicas: they compose notch-to-parameter mapping, full walks, round trips and degenerate inputs. They do not simulate a key event, which the reviewer noted and accepted given how thin the component wiring is.

leonfullxr:

leonfullxr:
Third opinion from `opencode-go/kimi-k3` (read-only, after Codex and CodeRabbit twice): **no high or medium**, two lows, a nit and one informational. Two taken.

The accessibility one is the reason this pass was worth running, because it lands on exactly what this PR is about:

**The thumb announced the notch index.** The track now measures parameter steps, and the thumb's `aria-valuemin/max/now` are the track's, so a screen reader on the realtime steps slider said "4 of 6" while the control read 6. The old fixed track announced a percentage, equally meaningless, so it is not a regression in kind, and a branch whose whole point is that the keyboard works for a narrow integer parameter should not leave the announced number wrong.

Fixing it needed the shadcn wrapper: `role="slider"` is on the thumb, so `aria-valuetext` on the root does nothing. `decisions.md` records that shadcn source is vendored and owned rather than imported, so the wrapper takes a `valueText` prop and passes it to the thumb. It is the same string already rendered beside the label, so the announced value and the visible one cannot drift.

Also taken: the comment cited 38 as the widest shipped notch count, which was the test fixture. The widest a manifest actually asks for is 49.

## Recorded rather than fixed

- **PageUp and PageDown do nothing.** bits-ui handles arrows, Home, End and Meta+Arrow only. Identical on the old track, unchanged here, and Home/End cover the jump case.
- **A declared strength minimum of 0.05 is unreachable** on `sdxl-base` and `vega-rt`: `inferStep` gives that range a step of 0.1 and `normToValue` rounds to it, so position zero sends 0.1. Byte-identical before and after this branch, and invisible because the label rounds too. Filed as #302.

## What it verified rather than assumed

- The rounding chain is exact for every shipped spec: one notch is one step, endpoints reachable, every notch round-trips, and the label cannot disagree with the payload because both come from the same `normToValue` call in the parent.
- Pointer drags snap to the same notches as the keyboard, so the reachable set is identical to the old track for every shipped spec.
- Model switching with a held value behaves as each panel already designed it, and this branch changes neither.
- The cap premise is real: bits-ui materialises one entry per notch and derives three arrays from it on every value change, even though this wrapper renders no tick marks.

`npm test` 60 passing, lint, `svelte-check` and the production build all clean.

Issue #281, opened 2026-08-13
A WebP header without its bitstream passes inspection


leonfullxr:
# Description

`_webp_dimensions` in `backend/app/storage.py` reads the RIFF container and the chunk headers and never decodes. A `VP8 ` chunk needs ten header bytes and a `VP8L` chunk five, so a 30-byte or 26-byte container carrying a header and no compressed data is accepted as a usable WebP: `image_info` reports `image/webp` with plausible dimensions, and `job_done` creates the thumbnail asset row.

# Motivation

The thumbnail inspection added in PR #277 exists to stop a worker having the studio serve arbitrary bytes as an image. It closes the case where the object is not a WebP at all, and not this one: a worker can still obtain a thumbnail row for a file no decoder will render. The result is a broken image in the gallery chosen by the worker, not active content, because the API serves it with `nosniff`.

Both readers are deliberately parse-only. `_png_dimensions` says why: an earlier version decompressed and that was twice a denial of service, and deciding how much memory to spend from a number the peer supplied cannot be made safe. A real decoder belongs off this path entirely, which is why this is filed rather than fixed inline.

# Functionality

Decide one of:

- Accept the limit. Document that inspection proves structure and not decodability, for both formats, and rely on `nosniff` plus the fact that a broken thumbnail is cosmetic.
- Add a decode step somewhere that is not the API request path: a bounded worker-side or out-of-process check whose failure drops the thumbnail row the way a failed inspection already does.

# Details

`backend/app/storage.py`, `_webp_dimensions` and `_inspect`; the consumer is the `job_done` branch of `on_worker_message` in `backend/app/jobs.py`.

Found by review of PR #277 (codex gpt-5.6-sol, xhigh), round 3.

Issue #278, opened 2026-08-13
A presigned upload URL outlives the object it was minted for


leonfullxr:
# Description

An S3 presigned PUT stays valid for its full hour. After the API has verified an object and cleanup has deleted it, the same URL can recreate it: `If-None-Match: *` only refuses a write when a current object exists, and a deleted key has none. The recreated object is referenced by nothing, so no later collector removes it.

# Motivation

PR #277 made outputs write-once so a verified object cannot be replaced. That closes the replacement case and not this one. `purge_attempt_blobs` runs on the failure and stall paths, and the success path deletes earlier attempts, so a worker that keeps its upload URL can put bytes back at any of those keys afterwards. The PUT carries no size condition, so what it leaves is bounded only by the worker.

The exposure is a compromised or buggy worker leaving persistent unreferenced billable objects in the bucket, not a path to another user data. The local backend does not have the same hole: its authority is re-checked against the live dispatch when the body arrives, and a purged attempt is no longer in flight.

# Functionality

Pick one of:

- A lifecycle rule that expires objects under the dispatch prefix, with the status cleared for keys a committed asset row names.
- A cleanup pass that runs after the upload URL can no longer be valid, rather than at the moment the attempt ends.
- A shorter TTL for upload URLs than for the browser-facing GET URLs, which shrinks the window without closing it.

# Details

`SIGNED_URL_TTL` in `backend/app/storage.py` is one hour and is shared by the upload target and the view URL. The bucket lifecycle belongs to the cloud infrastructure rather than to this repository, which is part of why this was kept out of #277.

Found by review of PR #277 (codex gpt-5.6-sol, xhigh).

leonfullxr:
Still open, and still cloud infra, not this repo. #328 pins Content-Type on the presigned GET and narrows the nosniff claim. It does not add an S3 lifecycle rule.

PR #297, opened 2026-08-14
Record the dispatch capability, and put the token in the blueprint

leonfullxr:
Documentation only.

`decisions.md` had no entry for what shipped in #277 and its follow-ups, so the reasoning lived in commit messages and a closed PR: why a dispatch mints a token, why the protocol moved to 3 rather than accepting a missing field from every worker, why outputs are written once, and why a thumbnail is now read rather than believed.

The entry also records the two deliberate limits, because both are filed issues rather than oversights:

- a WebP header with no bitstream still passes a parse-only reader (#281), and both readers stay parse-only because the PNG one that decompressed was twice a denial of service;
- a presigned S3 PUT outlives the object it was minted for, so a key that cleanup deleted can be recreated within the hour (#278), which wants a bucket lifecycle rule rather than application code.

`blueprint.md` is structural rather than normative for the wire (`connection-handling.md` is), but its worker protocol table named neither the token nor the upload targets, which made `dispatch_job` look like less than it is. One row changed, one paragraph added, with the pointer to the normative document.

Mermaid renders (35 diagrams), no non-ASCII.

## Summary by CodeRabbit

* **Security & Reliability**
  * Improved job tracking so responses from unrelated or outdated job attempts are ignored.
  * Added stronger authorization for local-storage uploads.
  * Prevented duplicate output writes across local and S3 storage.
  * Added validation for PNG and WebP thumbnail data.
* **Documentation**
  * Documented dispatch-token handling, worker protocol compatibility, storage protections, and validation behavior.


leonfullxr:
Review round 1 (codex `gpt-5.6-sol`, xhigh, read-only): five findings, three medium, all applied.

The review was worth running precisely because a decision entry that misdescribes its own subject is worse than none, and this one did:

| Finding | Correction |
|---|---|
| **Medium.** "Outputs are written once: `O_EXCL` locally" is false | The local route writes a temporary, re-checks authority, and publishes with `os.link`. `O_EXCL` was the first version and was replaced, because the key is readable while the body is still landing. Now described correctly, listed under rejected alternatives with the reason it lost, and the hard-link requirement on `STORAGE_LOCAL_PATH` is stated |
| **Medium.** `blueprint.md` still showed `upload_target(self, key)` in the Storage seam | Signature updated to `(key, token)` with a note that the token authorises the local PUT while S3 signs instead. As the reviewer put it, the stale seam could have a new storage implementation omit local upload authorisation entirely |
| **Medium.** "requiring real image data" overstates the WebP walk | It requires an image chunk with a valid frame header; a header with no bitstream still passes, which is #281, stated two paragraphs later. The overstatement undercut that issue's own brief |
| **Low.** "the whole of the N-1 exception" | A `Worker` built without a registration is lenient the same way until #282 lands. Now says so |
| **Low.** "identified by the job id alone" | It was the job id plus the worker object; the defect is that those two cannot separate attempts on the same worker |

Also confirmed: none of the five rejected alternatives is a straw man, the entry does not contradict `connection-handling.md` or `api.md`, and placement and voice fit its neighbours.

One structural note the reviewer raised and I agree with: the protocol 2/3 exception now appears in `decisions.md`, `blueprint.md` and `connection-handling.md`. It is consistent today and will drift when the floor moves. `connection-handling.md` is the normative one; the other two should point at it rather than restate the rule, which the blueprint paragraph already does.

Finally, it flagged one decision from #287 that deserves its own entry once that merges: holding the dispatch entry, the tracking state and the worker slot until the terminal commit is durable, and accepting a leaked slot if the process dies between commit and release. I will write that when #287 lands rather than pre-recording an unshipped choice. #282, #256, #280 and #250 do not need entries: they enforce or fix existing decisions rather than making new ones.

leonfullxr:

PR #299, opened 2026-08-17
Reach the fallback when a page of the anchor hunt fails

### LINKED ISSUE #251: A failed page during the viewport anchor search leaves the canvas uninitialised
# Description

If the Images canvas is restoring a saved viewport whose anchor root is not on the first page of
roots, it requests another page and returns without marking the viewport ready. When that request
fails, nothing runs the initialisation again, so the canvas stays uninitialised until the user
retries by hand.

# Motivation

`viewportReady` gates more than the first paint. The effect that follows the anchor through
repacking requires it, so a failure in the anchor search leaves the canvas in a state where later
successful loads still do not centre anything.

The existing root-failure affordance does recover it, because retrying eventually reaches a
successful page which schedules initialisation. So this is a stall rather than a permanent break.
It is still wrong: a transient network failure during the anchor search should not require the
user to notice and act.

# Functionality

`initializeViewport` should not depend on a load succeeding to make progress. Options, not a
decision:

- Treat a failed anchor-search page as exhausting the search, falling through to the newest-root
  fallback that already exists for an unresolvable anchor.
- Count failures against `MAX_ANCHOR_SEARCH_PAGES` so the search terminates on errors as well as
  on pages.
- Re-run initialisation from the failure path rather than only from the success path.

# Details

`frontend/src/lib/components/lineage-canvas.svelte`: the early return in `initializeViewport` that
increments `anchorSearchPages` and calls `loadRoots`, and the `if (loaded && !viewportReady)`
scheduling at the end of `loadRoots`, which only runs when a page succeeded.

This predates PR #239. It was introduced with the anchor-search cap in 38f5105 and is on main
today. Found by review of #239, filed separately because it is not a regression from that branch.

leonfullxr:
# Description

Closes #251. A failed page during the viewport anchor search left the Images canvas
uninitialised, with nothing able to schedule initialization again.

# Details

Restoring a saved viewport anchored to a root that is not on the first page requests another page
and returns without setting `viewportReady`. The scheduling at the end of `loadRoots` read
`if (loaded && !viewportReady)`, and `loaded` is set only in the success path, so a failed request
left the canvas stranded: the anchor-following effect requires `viewportReady`, so even a later
successful load centred nothing. The user saw an empty canvas and could only recover by noticing
the existing root-failure retry.

Two coordinated changes:

- The scheduling decision moves into `decideViewportScheduleAfterRootPage` beside the other canvas
  rules. `initialize` now wins whether or not the page arrived, so a failure reaches the
  newest-tree fallback. A failed page still never recentres, because centring against content that
  never arrived would land on stale roots.
- A failure ends the anchor hunt instead of spending the rest of `MAX_ANCHOR_SEARCH_PAGES` against
  an endpoint that is not answering. The existing retry clears `rootsFailed` and the hunt resumes.

The decision is extracted rather than left inline because this file has repeatedly regressed on
logic that lived only inside a component effect, and an extracted rule is the only shape that can
be guarded here.

# Verification

`npm run lint`, `npm run check` clean across 1247 files, 54 unit tests, and `npm run build` with
CSP verification all pass.

Both rules mutation-tested. Restoring the old condition, so `initialize` requires a loaded page,
fails the new test. Letting a failed page recentre also fails it. Neither rule is decoration.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lineage canvas viewport behavior after root data loads.
  * Prevented unnecessary recentering when root loading fails.
  * Improved handling of empty accounts and retry scenarios.
  * Stopped viewport pagination when root loading cannot continue.

* **Tests**
  * Added coverage for viewport initialization, recentering, no-op states, retries, and failed loads.


leonfullxr:
Confirmed and fixed in ff97de3. The window is real: `reloadRoots` increments `rootsRequestEpoch` and clears `roots`, so a frame scheduled just before a filter toggle still ran `initializeViewport`, which has no epoch check and ends with an unconditional `viewportReady = true`. It marked the viewport ready having centred nothing, and the replacement request then read that as ready and scheduled nothing.

The recenter callback had the same window and would have centred on roots from a superseded request.

Both callbacks now re-check through one `stillCurrent` closure, which also replaces the two existing copies of those four conditions in the same function.

Worth noting the independent reviewer checked whether a current request could fail the guard, concluded correctly that none could, and did not consider the gap between the guard and the callback. This one was yours.

PR #304, opened 2026-08-17
Say when a tree's derivatives could not be loaded

### LINKED ISSUE #235: A subtree that fails to load is indistinguishable from a root with no derivatives
# Description

When a subtree request fails permanently, the Images canvas shows nothing to say so. The tree
collapses to its single root node, which is visually identical to a root that genuinely has no
derivatives. Descendants are silently absent.

Confirmed in the browser on 2026-08-10 against main: with `/subtree` forced to return 500, root
`982d7094` (which has five real children) rendered as a lone tile, and the canvas contained no
text matching failure, error or retry anywhere.

# Motivation

This is the one failure mode of the canvas that misinforms rather than degrades. A user looking
at a root with five derivatives sees a root with none, and has no reason to doubt it. The whole
point of the section is that structure is visible, so silently dropping structure is worse than
showing an error.

It is also unrecoverable within the session. After the single automatic retry is spent,
`decideLineageTreeLoad` returns `skip` for the rest of the section's life, and failed entries are
not written to `sessionTreeCache`, so only leaving the section and returning clears it. Nothing
tells the user that leaving and returning is the remedy.

# Functionality

Give a failed tree an affordance of its own, in the spirit of the existing `rootsFailed` retry
already present for the root list:

- Mark the root tile as incomplete rather than childless.
- Offer a retry on that tile, which forces a load and grants a fresh retry budget, the same way
  `retainedRetryBudget(true, ...)` already treats a forced load.
- Keep it quiet at the constellation zoom band, where a per-tile control would not fit.

# Details

Relevant code in `frontend/src/lib/components/lineage-canvas.svelte`: `treeStatus === 'loading'`
renders a spinner, `'error'` renders nothing, and `basePackedTrees` falls back to the single-node
layout from `rootLayoutNode`.

The retry budget itself is correct and was verified in the same session: a persistently failing
subtree makes exactly two attempts, the initial load plus one retry, and stays at two across
repeated visibility passes. This issue is only about the absence of feedback.

Found during the review of PR #210 and confirmed by the browser pass afterwards.

leonfullxr:
# Description

Closes #235. A subtree that failed permanently collapsed to its root tile, which is visually
identical to a root that never had derivatives.

# Details

Confirmed in a browser when the issue was filed: with `/subtree` forced to 500, a root with five
real children rendered as a lone tile and the canvas contained no text matching failure, error or
retry. It was also unrecoverable within the session, because the one automatic retry is spent,
`decideLineageTreeLoad` then returns `skip` for the rest of the section's life, and a failed entry
is never written to `sessionTreeCache`, so only leaving the section and returning cleared it.

A failed root now carries a retry control that forces a load, which grants a fresh budget through
`retainedRetryBudget`.

Three details worth stating, because each was a constraint rather than a preference:

- The control is a **sibling** of the tile, not a child. The tile is itself a `<button>`, so
  nesting one inside it would be invalid, and the decorative `root-affordance` span already there
  takes no pointer events. This is the same shape `reset-tree-position` uses.
- It sits opposite `reset-tree-position` so a tree that was dragged and then failed shows both.
  They are mirrored about the tile centre and do not overlap: 196px apart at the card band, 84px
  at trees.
- It is hidden at the constellation band, where a 24px control would swamp a tile whose
  half-width is 18px.

The tile's `aria-label` gains the failure phrase too, so a screen reader is not told a tree is
complete when its derivatives are missing.

The retry budget itself is unchanged and was already correct. This issue was only ever about the
absence of feedback.

# Verification

`npm run lint`, `npm run check` clean across 1247 files, 54 unit tests, and `npm run build` with
CSP verification all pass. The `en.json` and `es.json` key sets are identical.

No unit test: the change is markup, CSS and an accessible name, and the state it keys off,
`treeStatus === 'error'`, is already covered. A test that rendered a fake component to assert on a
CSS class would be the theatre this repo has rejected before.

### DIFF
--- a/frontend/src/lib/components/lineage-canvas.svelte
+++ b/frontend/src/lib/components/lineage-canvas.svelte
@@ -40,6 +40,7 @@
 	import MoveIcon from '@lucide/svelte/icons/move';
 	import PencilIcon from '@lucide/svelte/icons/pencil';
 	import PlusIcon from '@lucide/svelte/icons/plus';
+	import RefreshCwIcon from '@lucide/svelte/icons/refresh-cw';
 	import RotateCcwIcon from '@lucide/svelte/icons/rotate-ccw';
 	import ScanLineIcon from '@lucide/svelte/icons/scan-line';
 	import StarIcon from '@lucide/svelte/icons/star';
@@ -1617,7 +1618,7 @@
 									scheduleTreeLoad(root, true);
 								}}
 							>
-								<RotateCcwIcon />
+								<RefreshCwIcon />
 							</button>
 						{/if}
 					{/if}
--- a/frontend/src/lib/components/lineage-canvas.svelte
+++ b/frontend/src/lib/components/lineage-canvas.svelte
@@ -1503,7 +1503,7 @@
 						class:is-missing={data.entry.missing || shownImage === null}
 						class:is-starred={starred}
 						style={`--tile-pull: ${proximityScale(item)}`}
-						aria-label={`${actionLabel(data.entry.action)}: ${promptLabel(data)}${starred ? `. ${t('app.images.starred')}` : ''}${item.isRoot ? `. ${t('app.images.drag_tree')}` : ''}${item.treeStatus === 'loading' ? `. ${t('app.images.tree_loading')}` : ''}`}
+						aria-label={`${actionLabel(data.entry.action)}: ${promptLabel(data)}${starred ? `. ${t('app.images.starred')}` : ''}${item.isRoot ? `. ${t('app.images.drag_tree')}` : ''}${item.treeStatus === 'loading' ? `. ${t('app.images.tree_loading')}` : ''}${item.treeStatus === 'error' ? `. ${t('app.images.tree_load_failed')}` : ''}`}
 						title={promptLabel(data)}
 						onfocus={() => (focusedNodeId = item.node.id)}
 						onblur={() => (focusedNodeId = null)}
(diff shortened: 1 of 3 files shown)


leonfullxr:
Not browser-verified. The Chrome extension is not connected in this session, so I could not drive the canvas and see the control in place.

What I did check, statically against the source: the button renders only for `item.isRoot && item.treeStatus === 'error'`, resolves the root through `persistedRoots`, does nothing when that lookup misses, stops `pointerdown` so it cannot start a tree drag, calls `scheduleTreeLoad(root, true)` so the forced load grants a fresh budget, is `display: none` at the constellation band, and the failure phrase reaches the tile's `aria-label`.

The two controls cannot overlap. Both are 24px, mirrored about the tile centre and anchored at `half-width - 10px`, so they sit 196px apart at the card band and 84px at trees, and the retry control is hidden at constellation where half-width is 18px.

What remains unverified is what it actually looks like: contrast against the tile, whether the icon reads as retry rather than reset given both use `RotateCcwIcon`, and the focus ring. Worth a look before merge by anyone with a browser to hand.

leonfullxr:
Independent review came back ready to merge, with three non-blocking notes. Acting on one, disputing one.

**Acted on: the two controls shared an icon.** Both used `RotateCcwIcon` on the same tile, distinguished only by corner and tooltip. Retry now uses `RefreshCwIcon` (d418392).

**Disputed: that the new Spanish string should drop its accent for file consistency.** It should not. `es.json` has 171 accented lines out of 549, including `Código abierto`, `Navegación principal` and `Imágenes generativas`. Accented Spanish is the file's convention; the unaccented neighbours near this key (`imagenes`, `arbol`, `posicion`) are the anomaly, and two of that group were already corrected in #239 for the same reason. Unifying downward would spread a spelling error rather than fix one. Left as `árbol`.

**Noted, not fixed: focus drops to the document body** when the retry button unmounts on activation, because the status flips to `loading`. The reviewer confirmed `reset-tree-position` behaves identically, so this is a pre-existing pattern rather than a regression from this change, and worth an accessibility pass across both rather than a one-off here.

PR #305, opened 2026-08-17
Fence what the wire can carry, and give settlement a table

leonfullxr:
# Description

Four review comments on #300 arrived four minutes before it merged and were not read before the merge. Three of them are real, and one of those makes part of the merged design unbuildable as written. This is the fix.

# Motivation

The lifecycle design said `control_generation` travels on every lifecycle control and on frames in both directions. The binary frame header is 17 bytes, one kind byte and one 16-byte session UUID, and [connection-handling.md](docs/connection-handling.md) already says it carries no sequence number or revision, with issue #19 owning the field that would add one. An implementer following the design would have had to change the wire format that another issue owns, or quietly skip the frame half of the fencing.

It also required the terminal transaction to commit an outbox record under a stable settlement key, retry it and deduplicate it, against a data model that has no outbox table. And it stored per-attempt counters without the classification `session_closed` carries, while [metrics.md](docs/metrics.md) promises a category for every closed realtime session.

# Functionality

Frames are not fenced by a field they cannot carry. A canvas frame lands on whichever runner the worker holds for that session, and only a fenced open installs one, so the real gap was a superseded runner still able to emit a frame the API would accept. Accepting an open now cancels and discards the runner it replaces, which closes the gap without touching the header. If #19 later adds a revision, the generation can ride there and frames become fenced directly rather than by inference.

`settlement_outbox` joins the data model, keyed by the settlement key rather than a surrogate id, because that key is what makes a retried delivery a no-op instead of a second charge. `realtime_session_attempts` gains `category` and `category_score`, so a crash after the attempt row commits can rebuild what a session made as well as what it used.

# Details

`docs/api.md` also described protocol 4 as though it had shipped. It now says what ships, the unfenced `update_session` at protocol 3, and marks the fenced form as designed. The sweep script's docstring says "three sessions" rather than an unqualified "three".

The two findings coderabbit raised at 09:16 about the attempt table, a composite key and cumulative `duration_ms`, were already fixed in #300 before the merge; they were reported against an earlier snapshot.

## Summary by CodeRabbit

* **Documentation**
  * Clarified realtime WebSocket protocol version support and planned lifecycle improvements.
  * Documented durable session settlement, delivery tracking, and recovery behavior.
  * Clarified connection handling, runner replacement, generation ordering, and session settlement.
  * Updated performance notes to distinguish measured batching improvements from unavailable caching.
* **Bug Fixes**
  * Improved stalled-connection diagnostics by distinguishing received frames from empty drain windows.
  * Preserved recent control messages and reported socket state for clearer troubleshooting.


leonfullxr:
@coderabbitai review

The merge-risk assessment above is against `119f5`, the first of nine commits. The three concerns it raises were addressed after it:

- Stale-frame delivery: the claim that supersession delivers nothing was wrong and is corrected. Cancelling a runner inside its GPU work delivers nothing, but one cancelled in its send may already have the bytes on the transport, so the stated bound is one already-produced frame and never a second, with frame-level fencing left to issue #19 rather than claimed.
- Duplicate settlement on retry: `session_checkpoint` and `session_closed` are one max-upsert on an attempt row keyed by session, generation and worker incarnation, idempotent because the totals are cumulative, and the outbox row is keyed by the settlement key so a retried delivery is a no-op.
- Incomplete classification: the attempt row now carries `category` and `category_score` for a crash after it commits, and for a worker that dies without closing the design states there is no category at all, because the label comes from classifying a final frame that was never sent.

Please review the branch as it stands.

leonfullxr:
@coderabbitai full review

PR #309, opened 2026-08-17
Retry the deletes the terminal paths could not make

### LINKED ISSUE #254: A failed cleanup delete is never retried
# Description

A failed cleanup delete is logged and forgotten. Nothing retries it, and no later collector finds the object.

# Motivation

`purge_attempt_blobs` in `backend/app/jobs.py` runs on every terminal path and swallows per-key failures so one bad key does not stop the rest, which is right. What is missing is what happens afterwards: the terminal state is already committed, the asset row names only the winning key, and no sweep looks for orphans. A delete that fails because of a denied IAM permission, a transient S3 error, or a full disk leaves the object permanently, silently.

The log line is now a warning with the exception, so the failure is visible to an operator reading logs. It is not visible to the system.

# Functionality

- Record keys whose deletion failed, durably enough to survive a restart.
- Retry them on a schedule, alongside the existing five-minute maintenance pass in `gpu_samples.maintain_loop` or as its own sweep.
- Give up after a bounded number of attempts and log once at error, so a permanently undeletable key does not retry forever.

# Details

An alternative worth weighing first: an S3 lifecycle rule expiring noncurrent versions and a prefix rule for orphaned attempt keys would cover this and every path nobody thought about, at no application cost. It only covers the cloud profile, so a self-hosted install on local storage would still need the retry, but local deletes fail far less often.

Low severity: it needs a delete to fail, and the common cause of that is a misconfiguration an operator will hit immediately and loudly rather than silently over time.

Found by adversarial review of PR #243.

leonfullxr:
Closes #254. Implemented by OpenCode (deepseek-v4-flash, max effort) to a design I fixed in the brief; gated and mutation-checked by me.

purge_attempt_blobs swallows per-key failures so one bad key does not stop the rest, and nothing retried them. The terminal state is already committed, the asset row names only the winning key, and no sweep looks for orphans, so a delete that failed on a denied permission or a transient error left the object forever: visible in a log line, invisible to the system.

Failures go into a pending_deletes table, one row per key. A sweep every five minutes retries them oldest first, a hundred per pass, backing off by doubling minutes to an hour. The eighth failure gives up, logs once at error and drops the row, because a row nothing will retry is a lie about the object.

Recording is best effort: the database can be down at the moment a delete fails, and a cleanup failure must never fail a job.

## The alternative, named rather than ignored

On S3 a lifecycle rule expiring noncurrent versions would cover this and every path nobody thought of, at no application cost. It covers only the cloud profile, and self-hosted local storage would still need this, so the retry lands first. The lifecycle rule is still worth having and belongs with #278.

## Verified here

Five new tests, each proved by breaking the line it covers: the recording call, the give-up threshold, the due-time filter, and the backoff all fail their test when removed.

Migration 0013 applied to a fresh database and rolled back: the table appears with the five expected columns and the downgrade drops it.

make verify green: 321 backend, 135 worker, frontend build and CSP.

## Summary by CodeRabbit

* **New Features**
  * Failed blob deletions are now tracked and retried automatically in the background.
  * Retries use controlled backoff, timeouts, attempt limits, and periodic maintenance.
  * Cleanup errors retain relevant details for monitoring and recovery.

* **Bug Fixes**
  * Temporary storage deletion failures can recover without manual intervention.
  * Local file deletion no longer blocks other application work.

* **Database**
  * Added persistent tracking for pending deletions, retry attempts, failure details, and retry timing.

### DIFF
--- a/backend/app/jobs.py
+++ b/backend/app/jobs.py
@@ -21,7 +21,7 @@ import unicodedata
 import uuid
 from collections.abc import AsyncIterator
 from dataclasses import dataclass
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from itertools import count
 from typing import Literal, Protocol
 
@@ -29,6 +29,7 @@ from fastapi import APIRouter, Depends, HTTPException, Query, Response
 from fastapi.responses import StreamingResponse
 from pydantic import BaseModel, ConfigDict, Field
 from sqlalchemy import and_, func, or_, select, text, update
+from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.engine import RowMapping
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import aliased
@@ -38,7 +39,7 @@ from app.auth import current_user, require_role
 from app.manifests import validate_params
 from app.settings import get_settings
 from app.storage import get_storage
-from app.tables import Asset, Job, User
+from app.tables import Asset, Job, PendingDelete, User
 
 logger = logging.getLogger("potocolom.jobs")
 
@@ -57,6 +58,12 @@ DOWNLOAD_SLUG_MAX_WORDS = 6
 LINEAGE_SUBTREE_MAX_DEPTH = 100
 # Six hundred matches the canvas tile ceiling and bounds unusually broad trees.
 LINEAGE_SUBTREE_MAX_NODES = 600
+# Sweep policy for blobs the terminal paths could not delete (issue #254).
+PENDING_DELETE_ERROR_MAX = 1000  # a last_error is a sentence, not a traceback
+PENDING_DELETE_PASS_LIMIT = 100  # rows per pass, so one bad batch cannot hog a tick
+PENDING_DELETE_MAX_ATTEMPTS = 8  # after this many failures the object stays
+PENDING_DELETE_BACKOFF_CAP = 60  # minutes; 2 ** 7 would already overshoot an hour
+MAINTAIN_DELETES_INTERVAL = 300.0  # seconds
 
 
 class Queues(Protocol):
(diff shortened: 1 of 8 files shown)


leonfullxr:
Review from `opencode-go/kimi-k3` (read-only): **not ready to merge**, one medium, one medium-low, two lows and a nit. All five taken in 4d33eca.

The medium is the one worth stating plainly, because it is the same bug this PR exists to kill:

**The success path leaks exactly the way `purge_attempt_blobs` did.** I wired the retry into the failure path and not into the success path, which collects the earlier attempts and an unreported thumbnail through its own `except` and swallowed the failure. One function away from the fix. It records now, and a test drives a denied thumbnail delete through a job that succeeds.

| Finding | Action |
|---|---|
| Medium-low: two replicas process the same due rows, and the loser's commit fails on rows the winner removed | `FOR UPDATE SKIP LOCKED` on the select. The cloud profile runs this loop in every replica on the same cadence |
| Low: giving up at eight attempts covers about three hours, then the leak is permanent and traceless | Giving up no longer drops the row. `next_attempt_at` becomes null, which the sweep skips, and the row stays as the only record that the object exists. The migration makes the column nullable |
| Low: no timeout around the storage call | Bounded. `LocalStorage.delete` is an unlink, which blocks forever on a wedged mount: the pass would never finish, the loop would never sleep, and nothing would raise |
| Nit: the model docstring said non-sweep writers only refresh `last_error` | Corrected, and it now explains what a null `next_attempt_at` means |

One thing I dropped while applying this: my first version also added `.where(next_attempt_at.is_not(None))`, and the mutation check showed removing it changed nothing, because `null <= now` is null and the existing comparison already skips those rows. A line no test can hold is a line that will drift, so it is a comment instead.

What kimi checked and cleared, so it does not need re-deriving: a crash mid-pass is safe because both backends delete idempotently; recording cannot disturb the caller's transaction, since both callers invoke it after their terminal commit; the primary key is the right upsert target and the missing index on `next_attempt_at` is fine for a table this size; and the sweep cannot delete a key an asset row will later name, because attempt numbers never repeat and asset rows only ever name the winning attempt.

`make verify` green: 322 backend, 135 worker.

**Codex still owes this branch a round** and is out of quota; I will run it when the limit resets rather than merging on one reviewer.

leonfullxr:
Second review, by a Claude opus xhigh instance, since codex is out of quota until Aug 20 and the standing rule is that the review does not disappear because the usual reviewer is down. **Not ready to merge**, one medium-high, two medium, one low-medium, one low. All five taken.

**The medium-high is a good catch against my own comment.** I wrapped the sweep's delete in `wait_for` and wrote that it bounds a wedged mount. It did not: `LocalStorage.delete` never awaited, and `wait_for` cannot interrupt a coroutine that never suspends, so the blocking `unlink` would have stalled the whole event loop, every socket and request in the process, with nothing raising to say so. The unlink is off the loop now, and a test runs a task alongside a delete to prove the loop keeps turning. The comment asserting the protection is exactly what would have stopped the next reader from checking.

| Finding | Action |
|---|---|
| Medium: one transaction held across every storage call | One row per transaction. A hundred rows against slow storage held a row lock and a pooled connection for the whole pass, a crash lost every counted attempt, and a terminal path recording a failure for a locked key waited behind it |
| Medium: giving up was a dead end nothing re-arms | Removed. The eighth failure still alerts once; the row stays scheduled. An object nobody can delete costs one call an hour and keeps a row saying so |
| Medium (same finding): `now` stamped once per pass | Stamped per row. A slow pass was handing its tail a due time already in the past, so the backoff silently vanished |
| Low-medium: no index, and given-up rows accumulated | Index in the migration, and with give-up gone the table no longer holds rows nothing will ever remove. The model now matches the migration on the `attempts` default |
| Low: `_trim_error` can store an empty string | The type name goes in front. `str()` is empty for `TimeoutError` and several botocore and OS errors, and this string is the whole record of why an object is still there |

The give-up removal is a deliberate deviation from the issue text, which asked for a bounded give-up with one error log. Both reviewers landed on the same objection from different directions: kimi said dropping the row destroys the only record of the object, and this round said unscheduling it makes an outage longer than the backoff permanent. Retrying hourly forever is smaller code than either and keeps the record; the alert at eight attempts stays.

It also confirmed what the earlier round found clean, so that does not need re-deriving: recording cannot disturb the caller's transaction, because both call sites run after their session has committed.

`make verify` green: 323 backend, 135 worker. **Codex still owes this branch a round on Aug 20**; two independent reviewers have now read it, so I am content to leave it open until then rather than merge on the fallback alone.

leonfullxr:
Third review, `opencode-go/kimi-k3`, now the reviewer of record: two medium, seven low, two nits. Both mediums and most of the lows taken in 9061cae.

**Both mediums are the same wedged-mount scenario the previous commit was built for, and neither was closed by it.** The unlink went off the loop, so the loop survived; the thread did not. `wait_for` abandons the await and the thread stays stuck for the life of the process, so against the default executor that is one leaked thread per attempt, hourly, forever now that nothing gives up, until every other off-loop read in the app starves behind it. Deletes have their own four-thread pool now. The second: the two terminal paths had no bound at all, so the same hang simply moved from the event loop, where it was loud, to one task and one thread, where it is silent. Every delete goes through one bounded helper.

| Low | Action |
|---|---|
| The recording upsert wrote its own due time, undoing a backoff the sweep had just set while it waited on that row lock | Writes `last_error` only, which is what the model docstring always claimed. New test holds it |
| A null check on a NOT NULL column, left from the earlier shape | Removed |
| The model named its index differently from the migration, so `create_all` and migrated schemas disagreed | Declared with the migration's name |
| The alert-once test passed against an alert-every-time mutation | Drives a ninth failure and counts messages |
| One test did not clear the table first | Cleared |
| No jitter, so replicas run the loop in phase forever | Jittered |
| The give-up reversal was undocumented | `decisions.md` entry, with the two rejected shapes and why the S3 lifecycle rule belongs with #278 |

**Not taken:** clock skew between replicas (both due checks use the app clock, so a replica running minutes ahead can retry a key inside another's backoff window). Real, but the cost is one duplicate delete against an idempotent operation, and moving both stamps to the database clock is a wider change than this branch should carry. Worth its own issue if it ever shows up in logs.

kimi also confirmed the two-phase sweep shape is sound under the interleavings I asked about: two replicas cannot double-retry, a row re-recorded while the sweep holds it converges either way, an overrunning pass delays rather than overlaps, and a crash between delete and commit is safe because both backends delete idempotently.

`make verify` green: 324 backend, 135 worker. This is the third round on this branch; the loop ends when a round returns nothing medium or higher, so it wants one more.

leonfullxr:
Round four, `opencode-go/kimi-k3`: **mergeable**, no correctness bug in the new machinery. Two lows and one pre-existing hole. That ends the loop: four rounds, and this one returned nothing medium or higher.

Taken: the orphan-path test read its row straight after polling for `succeeded`, and the commit that publishes that state happens before the orphan delete and its recording, so on a loaded loop the test could win the race. Both recording tests share one deadline helper now.

Filed rather than fixed here:

- **#313**, the verdict handler parking for two to four minutes on a wedged mount. Each delete is bounded at sixty seconds and there can be four of them, and the fleet loop awaits the handler inline, so `worker.last_seen` goes stale and `reap_dead_workers` can reap a healthy worker. kimi called this optional hardening, since it only matters once storage is already degraded; the fix is to hand the cleanup to a background task after the commit, with the sweep as the backstop it already is.
- **#312**, and this one predates the branch: a job that fails **without** a verdict never has its blobs collected at all. `sweep_stalled_jobs`, `on_worker_lost` and `recover` all funnel into `requeue_or_fail`, which calls `mark_failed` and nothing else, while the only two callers of `purge_attempt_blobs` are the worker verdicts. Attempt 2 uploads, the worker dies before `job_done`, the sweeper fails the row, and the object is never named again. Same silent leak this branch removes, through a door it does not cover.

What round four cleared explicitly, so it does not need re-deriving: the dedicated pool is sound for shutdown and for tests that build many loops, and four threads is the right bound because it caps leaked threads rather than throughput (the sweep deletes one row at a time); nothing inside the branch can lose an object silently; and it tried to break the other six tests by reading and could not.

`make verify` green: 324 backend, 135 worker. Merging.

PR #310, opened 2026-08-18
Make a realtime model resident before its session renders

### LINKED ISSUE #270: A realtime session on a non-resident model renders nothing, silently and permanently
# Description

A realtime session can render nothing at all, silently, for the lifetime of the worker process.

`frame` raises when the model's rung is not `full`, and the rung comes from `_pick_rung`, which caches its answer. The eviction that exists to make room for a full-resident model, `_ensure_vram`, is only reached later through `_pipeline`, so it never runs for a model whose rung was already decided against it. Nothing clears the cached rung for a model that never loaded.

The sequence: the worker advertises two realtime models, a queued job on some third model consumes VRAM, then a canvas session opens on the second realtime model. `_pick_rung` sees the reduced free VRAM, answers `model_offload`, caches that, and every frame afterwards raises. `SessionRunner._run` catches the exception, logs it, and continues, so the browser holds an Active session with its sent-frame count climbing and nothing ever coming back. No error reaches the user and none reaches the API.

Before the realtime picker there was only one realtime model in the studio and warmup pinned it `full` at boot, so the check never bit.

# Motivation

The failure is silent on every surface that matters: the browser looks connected, the API sees a healthy session, and only the worker log carries the exception. A user cannot tell this from a slow model, and an operator cannot tell it from an idle one.

# Functionality

A session that cannot be served must say so, and a rung decision must not be permanent:

- Opening a realtime session for a model that is not full-resident either evicts to make it so, using the same path a queued job would take, or refuses the session with the existing refusal close code so the browser can say why.
- A cached rung is invalidated when the conditions it was computed under change, at minimum when a model is unloaded or evicted.
- A frame that raises repeatedly for the same session ends the session rather than looping: the current `except Exception: continue` turns a permanent failure into an invisible one.

# Details

`_pick_rung`, `_ensure_vram`, `_pipeline` and the `frame` guard are in worker/worker/engine.py; the swallowing loop is `SessionRunner._run` in worker/worker/client.py. The memory ladder and its rungs are described in docs/architecture.md, and the realtime session lifecycle and close codes in docs/connection-handling.md and docs/api.md.

Related: issue #264 covers a slot that is never released, which is the same class of problem seen from the API side.

leonfullxr:
# Description

Addresses the residency half of #270. A canvas session on a model that was not fully resident
rendered nothing at all, silently, for the life of the worker process.

Does not close #270: the refusal half is still open, see below.

# Details

`frame` refuses a model on an offload rung. The rung is cached per model by `_pick_rung`, and the
eviction that would free the VRAM lives on the load path, which `frame` never reaches once the
cached answer says offload. A queued job on another model was therefore enough to poison a
realtime session: it opened, every frame raised, the runner logged and continued, and the browser
held an Active session with its sent-frame count climbing and nothing coming back.

`ensure_realtime_resident` decides residency once, when the session's task starts.

- **It reuses `load_model`** rather than adding a second eviction. That path already evicts,
  re-selects the rung against the freed VRAM, and loads, all under the GPU lock.
- **It runs in the runner's own task, not the `open_session` handler.** The control loop reads
  every heartbeat and every frame for every session, so awaiting a model load there would stall
  the fleet socket until the load finished, which the API reads as a dead worker.
- **A pinned memory mode is left alone.** A pin is an operator's explicit choice, consistent with
  the existing rule that pins are never demoted. Such a session still gets the per-frame refusal.
- **The `frame` guard stays** as the safety net for a model evicted mid-session, and a model that
  still cannot be made resident is logged once rather than on every frame, where the noise buried
  the cause.

The third part of the issue, invalidating a cached rung when conditions change, was already done:
`_forget_rung_if_unloaded` exists and is called by `_drop_pipeline`.

# Still open on #270

Refusing an unservable session so the browser can say why. There is no per-session refusal on the
wire: the only close code is `CLOSE_PROTOCOL_VIOLATION`, which closes the whole fleet socket, and
`session_closed` sent unasked is deliberately ignored by `realtime.py`, which only settles ids
already in `closing_sessions`. That is a protocol decision rather than a local fix, and it is
written up on the issue with two candidate shapes.

# Verification

`ruff`, `mypy` over 12 source files, and 140 worker tests pass, up from 137.

Three new tests, each mutation-checked by breaking its rule and confirming the suite fails:

| Reverted rule | Failing test |
|---|---|
| the already-full early return | `test_ensure_realtime_resident_returns_true_without_evicting_when_full` |
| the pinned-mode guard | `test_ensure_realtime_resident_leaves_a_pinned_mode_alone` |
| the runner's residency call | `test_session_runner_logs_a_non_resident_model_once_and_keeps_rendering` |

## Summary by CodeRabbit

* **Reliability**
  * Improved realtime rendering when models are not fully resident in memory.
  * The system now attempts to restore model residency before processing frames.
  * Rendering continues with a single warning if full residency cannot be restored.
  * Resident models avoid unnecessary reloads or memory changes.
* **Tests**
  * Added coverage for resident, offloaded, stale, and pinned-memory scenarios.

### DIFF
--- a/worker/tests/test_client.py
+++ b/worker/tests/test_client.py
@@ -994,3 +994,34 @@ def test_an_api_that_sends_no_dispatch_token_gets_none_back():
     reports = [json.loads(m) for m in socket.sent]
     assert reports
     assert not any("dispatch_token" in r for r in reports)
+
+
+def test_session_runner_survives_a_residency_load_failure(caplog):
+    # A load that fails outside the rung ladder, missing weights or a bad
+    # import, must not take the session task with it. Dying here would leave
+    # the session open with no runner behind it and nothing logged after that
+    # point, which is quieter than the per-frame refusal it replaced. The task
+    # has to reach the frame loop so the frame guard still speaks.
+    socket = FakeSocket()
+    engine = RecordingEngine()
+
+    async def boom(_manifest):
+        raise RuntimeError("weights missing")
+
+    engine.ensure_realtime_resident = boom
+    manifest = _realtime_manifest("vega-rt", steps_default=4)
+    session_id = uuid.uuid4()
+
+    async def scenario():
+        runner = SessionRunner(session_id, socket, engine, manifest,
+                               ensure_seed(manifest.with_defaults({"prompt": "x"})))
+        runner.submit(b"first")
+        await asyncio.sleep(0.03)
+        runner.close()
+
+    with caplog.at_level("WARNING"):
+        asyncio.run(scenario())
+
+    # Reached the frame loop rather than dying at the residency call.
+    assert len(socket.sent) == 1
+    assert "could not make model vega-rt resident" in caplog.text
--- a/worker/tests/test_client.py
+++ b/worker/tests/test_client.py
@@ -579,6 +579,58 @@ def test_session_runner_observes_each_rendered_frame_for_its_model():
     assert len(socket.sent) == 2
 
 
+def test_session_runner_logs_a_non_resident_model_once_and_keeps_rendering(caplog):
+    # The residency decision happens once, before the frame loop, and a False
+    # answer is logged once: the frames that follow would log an exception on
+    # every draw, and the per-frame noise buried the cause. The session stays
+    # up either way, so the frame fallback still renders here.
+    socket = FakeSocket()
+    engine = RecordingEngine()
+
+    async def not_resident(_manifest):
+        return False
+
+    engine.ensure_realtime_resident = not_resident
+    manifest = _realtime_manifest("vega-rt", steps_default=4)
+    session_id = uuid.uuid4()
+
+    async def scenario():
+        runner = SessionRunner(session_id, socket, engine, manifest,
+                               ensure_seed(manifest.with_defaults({"prompt": "x"})))
+        runner.submit(b"first")
+        await asyncio.sleep(0.03)
+        runner.submit(b"second")
+        await asyncio.sleep(0.03)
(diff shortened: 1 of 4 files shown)


leonfullxr:
Review came back ready to merge with two follow-ups. Acted on one, since it was a regression I introduced.

**Fixed in 761bc08.** `ensure_realtime_resident` can raise for a reason the rung ladder does not cover, missing weights or a bad import, and that exception propagated out of the session task before the frame loop started. Runner tasks are only gathered at shutdown (`client.py:609`), so the task died with nothing logged and the session stayed open with no runner behind it. That is quieter than the per-frame refusal it replaced, which is the wrong direction for a bug whose entire complaint is silence. The failure is now logged and the task continues into the frame loop, where the existing guard still refuses the model. Mutation-checked: narrowing the `except` so the exception escapes again fails the new test.

**Not fixed, and it belongs to the other half.** A later load, a queued job or a second realtime session whose model cannot coexist in VRAM, can evict after `ensure_realtime_resident` returned True, leaving a cached `full` rung with no pipeline behind it. The reviewer confirmed this diff neither introduces nor worsens it. It is the same silent-session shape this issue is about, and it wants the refusal protocol rather than another eviction, so it stays with the deferred half.

On the eviction question I raised: confirmed acceptable. `load_model` holds the GPU lock, so it cannot interrupt a queued job's own load, and an evicted model is simply reloaded by that job's load path, which behaved that way before this change too.

141 worker tests now, ruff and mypy clean.

PR #311, opened 2026-08-18
Encode a realtime session's prompt once, not once per frame

### LINKED ISSUE #301: Cache a realtime session's prompt embeddings instead of re-encoding them every frame
# Description

A realtime session re-encodes its prompt on every frame. The prompt changes only when the user types; the canvas changes constantly. Encoding the prompt once per session and reusing the tensors removes 26 to 33 ms from every frame, and the frame it produces is bit identical.

# Motivation

Measured on the reference RX 7600 XT at 512 px, complete frames, 40 timed frames per point, comparing a string handed to the pipeline against `prompt_embeds` and `pooled_prompt_embeds` computed once:

| Model | Steps | Sessions | String each frame | Cached | Saving |
|---|---|---|---|---|---|
| sdxl-turbo | 1 | 1 | 165.5 ms | 139.5 ms | 25.9 ms (15.7%) |
| sdxl-turbo | 1 | 3 batched | 372.1 ms | 340.5 ms | 31.6 ms (8.5%) |
| vega-rt | 4 | 1 | 233.9 ms | 205.1 ms | 28.9 ms (12.3%) |
| vega-rt | 4 | 3 batched | 514.3 ms | 481.0 ms | 33.2 ms (6.5%) |

The maximum absolute difference between a cached-embedding latent and a re-encoded one at the same seed is exactly 0.0, because they are the same tensors. So this is not a quality trade in the way the distilled decoder (issue #214) is one; it is the same computation done once instead of many times.

It is worth more than cross-session batching (issue #294), which measures at about seventeen percent, and it composes with both. It also changes capacity: `vega-rt` at the four steps its manifest ships goes from missing the bar at three sessions to 490.3 ms, inside it.

# Functionality

The engine keeps the encoded prompt for a live realtime session and reuses it for every frame until something invalidates it. The cache key is the text that was encoded, so a prompt the user edits misses and re-encodes once, and `update_session` carrying a new prompt is exactly that case. Nothing else about a frame invalidates it: seed, canvas and structure strength do not enter the text encoders.

The cache belongs to the session rather than the engine. A cached rung and a calibrated slot count already outlive what they describe (docs/decisions.md), and an embedding held past the session that owns it would be the same mistake with a much larger tensor.

Memory is small but not free: the SDXL pair is roughly 2 x 77 x 2048 in fp16 for the sequence embeddings plus the pooled vector, a few megabytes per session. It should be released when the session ends, and it must be per session rather than global so two sessions with different prompts cannot collide.

# Details

`worker/worker/engine.py` currently passes `prompt` into the pipeline on every realtime frame, and `_realtime_prompt_kwargs` only computes embeddings itself for prompts long enough to need chunking. The change is to make that path always produce embeddings and to memoise them for the session, which also makes the long-prompt path free rather than expensive.

Verify by measuring: the p95 saving above should reproduce with `scripts/prototype-batch-sweep.py`, and a test should pin that two frames of one session at the same seed are identical whether the embeddings were cached or not, and that editing the prompt changes the frame.

Recorded in docs/decisions.md, "Realtime concurrency comes from one GPU serving several sessions, by decode first and batching second".

leonfullxr:
# Description

Closes #301. A realtime session re-encoded its prompt on every frame. The prompt changes when the
user types; the canvas changes constantly.

# Details

A cache existed, but it held one entry on the pipeline keyed by
`(manifest.id, prompt, negative_prompt)`, so two concurrent sessions with different prompts evicted
each other on every frame. That is precisely the case the issue is about. Short prompts were never
cached at all: `if chunk_count == 1: return prompt_kwargs` meant only a prompt long enough to need
chunking took the encoding path.

**Ownership was the main design question.** The engine has no session concept, so a per-session
cache could have been a dict keyed by session id. It is not: that needs explicit release plumbing,
and a missed release leaks a multi-megabyte tensor pair. The issue makes the same argument, noting
the rung cache and calibrated slot count already outlive what they describe. Instead the runner
creates a `PromptCache` holder and passes it to each `frame` call, so the entry dies with the
session and there is no release path to forget.

**SD3 keeps the string path.** It needs joint CLIP and T5 embeddings and the CLIP-only chunker
cannot satisfy that contract, so "always produce embeddings" is not literal. That branch and its
comment are untouched.

# Verification

`ruff`, `mypy` over 12 source files, and 145 worker tests pass, up from 141.

Five new tests. Two mutation-checked against the rules that matter:

| Reverted rule | Result |
|---|---|
| the single-chunk early return | 4 tests fail |
| the pipeline-level cache, so sessions collide again | `test_two_sessions_with_different_prompts_do_not_evict_each_other` fails |

The second is the one worth reading: restoring the old cache fails exactly the test for the defect
this change exists to fix.

# What is not verified here

**The timing.** The 26 to 33 ms per frame was measured on a GPU. There is none available to this
change, `scripts/prototype-batch-sweep.py` was not run, and no performance claim is made beyond
the issue's own measurement.

**Equivalence of the chunker's single-chunk output with diffusers' own encoding.** Removing the
early return means a short prompt now goes through the chunker rather than being encoded inside
the pipeline. The issue reports a maximum absolute difference of exactly 0.0 between a
cached-embedding latent and a re-encoded one at the same seed, which is the relevant measurement,
but it was taken on a GPU and the tests here use a fake encoder, so they pin call counts and
identity rather than numerical equivalence. Worth a GPU check before this is relied on for output
stability.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved realtime frame generation by reusing prompt processing across frames within a session.
  * Reduced repeated prompt work while maintaining separate processing for different sessions.
  * Preserved prompt handling during retry scenarios and across supported generation modes.

### DIFF
--- a/worker/tests/test_client.py
+++ b/worker/tests/test_client.py
@@ -18,7 +18,7 @@ from worker.client import (
     serve_connection,
     warmup_realtime,
 )
-from worker.engine import GeneratedFrame, SimulatedEngine
+from worker.engine import GeneratedFrame, PromptCache, SimulatedEngine
 from worker.manifests import SIMULATED_MANIFEST, Manifest
 from worker.settings import Settings
 
@@ -542,7 +542,7 @@ class RecordingEngine(SimulatedEngine):
         super().__init__(0.01)
         self.observed = []
 
-    async def frame(self, manifest, params, payload):
+    async def frame(self, manifest, params, payload, *, prompt_cache=None):
         await asyncio.sleep(0.01)
         return GeneratedFrame(payload, 200)
 
@@ -577,6 +577,40 @@ def test_session_runner_observes_each_rendered_frame_for_its_model():
     asyncio.run(scenario())
     assert engine.observed == [("vega-rt", 200), ("vega-rt", 200)]
     assert len(socket.sent) == 2
+
+
+def test_session_runner_passes_the_same_prompt_cache_to_every_frame():
+    """The prompt embedding cache is owned by the session: the runner must
+    hand every frame the same holder, so concurrent sessions each encode
+    their own prompt once instead of evicting each other (issue #301)."""
+    socket = FakeSocket()
+    received: list = []
+
+    class CacheRecordingEngine(SimulatedEngine):
+        def __init__(self):
+            super().__init__(0.01)
+
+        async def frame(self, manifest, params, payload, *, prompt_cache=None):
+            received.append(prompt_cache)
+            await asyncio.sleep(0.01)
+            return GeneratedFrame(payload, 200)
+
+    engine = CacheRecordingEngine()
+    manifest = _realtime_manifest("vega-rt", steps_default=4)
+
+    async def scenario():
+        runner = SessionRunner(uuid.uuid4(), socket, engine, manifest,
+                               ensure_seed(manifest.with_defaults({"prompt": "x"})))
+        runner.submit(b"first")
+        await asyncio.sleep(0.03)
+        runner.submit(b"second")
+        await asyncio.sleep(0.03)
+        runner.close()
+
+    asyncio.run(scenario())
+    assert len(received) == 2
+    assert received[0] is received[1]
+    assert isinstance(received[0], PromptCache)
 
 
 def test_session_runner_logs_a_non_resident_model_once_and_keeps_rendering(caplog):
--- a/worker/tests/test_engine.py
+++ b/worker/tests/test_engine.py
@@ -13,6 +13,7 @@ from worker.engine import (
     GeneratedFrame,
     OBSERVED_FRAME_SAMPLES,
     OBSERVED_FRAME_WINDOW,
+    PromptCache,
     REALTIME_SIZE,
     SimulatedEngine,
     _canvas_to_sketch_map,
@@ -29,6 +30,8 @@ class _FakeTensor:
 
 
 class _FakeTorch:
+    OutOfMemoryError = type("OutOfMemoryError", (Exception,), {})
+
     class _NoGrad:
         def __enter__(self):
             return None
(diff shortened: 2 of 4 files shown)


leonfullxr:
Review came back "ready to merge once finding 1 is closed": whether the chunker's single-chunk output matches diffusers' own `encode_prompt`, since removing the early return routes short prompts through the chunker for the first time. It named four conditions. Three are statically checkable and the fourth is settled by the manifests, so here is the evidence rather than a GPU run.

**(a) Padded to the encoder window with the pad token.** `_encode_clip_chunks` builds `input_ids = [bos, *content, eos]`, then `padding = window - len(input_ids)` and `input_ids += [pad_token_id] * padding`, with `pad_token_id` falling back to `eos_token_id` when the tokenizer has none. That is `padding="max_length"` by hand, provided `window` equals the tokenizer's `model_max_length`.

**The window is manifest-declared, which was the real risk.** Every realtime-capable model declares 77, which is CLIP's `model_max_length`:

| model | prompt_token_limit |
|---|---|
| sd-turbo | 77 |
| sdxl-turbo | 77 |
| vega-rt | 77 |

The only manifest declaring anything else is `sd35-medium` at 256, and it is `text_to_image` only, so it never reaches `frame`, and SD3 takes the string early return regardless.

**(b) No attention mask to the CLIP encoders.** The mask is passed only under `if getattr(text_encoder.config, "use_attention_mask", False)`, which is false for SD and SDXL CLIP, so diffusers' behaviour is matched rather than diverged from.

**(c) Sequence embeds from `hidden_states[-2]`, pooled from the projection.** Sequence embeddings are `encoder_output.hidden_states[-2]`. Pooled is `encoder_output[0]`, which for `CLIPTextModelWithProjection` is `text_embeds`, the post-projection vector, matching diffusers' `prompt_embeds[0]` for SDXL.

**(d) BOS and EOS from the tokenizer.** `tokenizer.bos_token_id` and `tokenizer.eos_token_id`, not literals.

So the four conditions hold by construction for every model that can take this path. What this does not do is compare the tensors numerically on real weights: the issue reports a maximum absolute difference of exactly 0.0 at the same seed, measured on a GPU, and that measurement remains the numeric evidence. A `torch.allclose` check against `encode_prompt` on a real model would be conclusive and is worth doing when a GPU is next in front of someone.

The other six findings came back clean, including that a cached entry cannot survive into different weights (`manifest.id` is in the key) and that the OOM retry is correct because the entry write is the final statement of the encode.

Issue #288, opened 2026-08-13
The advertised frame p95 measures a narrower region once observations take over


leonfullxr:
# Description

Two numbers are treated as the same quantity and are not. Calibration times the whole `_frame` call, while a session's observed timing starts immediately before the diffusion pipeline call, so it omits work that happens inside the same GPU lock: pipeline lookup, any load the frame triggers, and prompt encoding. After twenty observed frames the narrower number supersedes the calibration one for that model.

`GeneratedFrame.gpu_ms` is set in `_frame` in worker/worker/engine.py; `_calibrate_realtime` in the same file times the call around it; both end up in `_realtime_p95_ms`, which `realtime_p95_ms` reports and the heartbeat carries as `frame_p95_ms`.

# Motivation

The number is displayed and it is used. `/api/v1/models` carries it, the realtime picker labels each model with it, and any capacity derived from it inherits the narrowing. So a model's advertised cost can drop after twenty frames without the model getting faster, and a capacity calculation reading it is optimistic by however much serialised work sits outside the measured region.

A concrete way to see the gap: a session that changes a long prompt every frame forces text-encoder work under the lock before each timed diffusion call, and none of it appears in the number.

The picker's label was introduced to answer "what will this cost me", so an unstated change in what is being measured undermines exactly the thing it exists for.

# Functionality

Decide what the number means, then make both producers agree:

- Time the same region in both places. The GPU lock's occupancy is the quantity that matters for anything about capacity, since that is what other sessions wait on, so measuring from lock acquisition to release is the honest choice and makes the two directly comparable.
- Or keep two quantities and stop conflating them: label the picker with the calibration figure and use observations only for trend, in which case `realtime_p95_ms` must not silently switch source.

Whichever is chosen, the documents that describe the field should say which region it covers. `docs/connection-handling.md` documents `frame_p95_ms` on the heartbeat and `docs/api.md` documents `realtime_p95_ms` on the model list; neither says what the boundaries are.

# Details

Related, and the reason this surfaced: #285, where four attempts to derive realtime capacity from this number were reviewed and rejected. One of those reviews measured the discrepancy directly. This issue is narrower and stands on its own: even with capacity left alone, the label should mean one thing.

Note the observation path also filters to frames rendered at the manifest's declared defaults, which is deliberate and documented in decisions.md, so the two numbers differ in what they include as well as where they start and stop.

PR #314, opened 2026-08-18
Encode a realtime session's prompt once instead of every frame

### LINKED ISSUE #301: Cache a realtime session's prompt embeddings instead of re-encoding them every frame

leonfullxr:
# Description

Closes #301. A realtime session encodes its prompt once and reuses the tensors, instead of re-encoding the same text on every frame while only the canvas changes.

# Motivation

The cache that already existed does not cover the ordinary case. For a prompt short enough not to need chunking, `_prompt_kwargs` returns `{"prompt": prompt}` and returns early, so a cache hit hands back that same string and the pipeline's own `encode_prompt` runs again on the next frame. Only a prompt long enough to be chunked was genuinely cached, which is the case the existing comment was written for.

Measured on the shipped path rather than on a prototype, `DiffusersEngine.frame` for `sdxl-turbo` at its one step, 30 frames, reference RX 7600 XT:

| | p50 | p95 |
|---|---|---|
| Encoding every frame | 288.1 ms | 290.7 ms |
| Encoded once per session | 264.3 ms | 269.9 ms |

About 21 to 24 ms, with the output unchanged because they are the same tensors.

That is 7 to 8 percent here, not the 15.7 percent the issue claims. The issue measured a 165 ms prototype frame, and a shipped frame is 270 ms because the full VAE decode is still in it, so the same 24 ms becomes a much larger share once the distilled decoder lands (#214). The absolute saving is the number that holds either way, and the issue has been corrected.

# Functionality

`_realtime_prompt_kwargs` is used by the two realtime call sites in `_frame`. It delegates to `_prompt_kwargs` first and returns that result unchanged when it already carries `prompt_embeds`, so a chunked long prompt is never encoded twice. Otherwise it encodes once through the pipeline's own `encode_prompt` with classifier-free guidance off, since the realtime path fixes guidance at zero, and caches the result on the pipeline keyed by model, prompt and negative prompt. The cache lives on the pipeline for the same reason the existing one does: a model switch drops it with the weights rather than pinning embeddings for weights no longer loaded. The negative prompt is never encoded, because guidance is zero, but it still keys the cache so two prompts differing only in negative text cannot share an entry once guidance becomes settable.

An SDXL-class pipeline gets `pooled_prompt_embeds` as well; an SD1.5-class pipeline gets only `prompt_embeds`, because its `encode_prompt` returns two values and it does not accept a pooled argument. An SD3-class pipeline still defers to diffusers, which needs joint CLIP and T5 embeddings.

The job paths are deliberately untouched. They share `_prompt_kwargs` and pass a settable `guidance`, so embeddings encoded without classifier-free guidance would be wrong for any job asking for guidance above zero, and a job's prompt is new every time, so a cache buys it nothing.

# Details

The tests drive `frame()` rather than the helper. Seven tests of the helper alone all passed with both call sites reverted to `_prompt_kwargs`, leaving the helper in place and unused: a suite that certifies a no-op. The two tests that drive `frame()` fail in exactly that case, one per call site, and one of them also pins that two frames with the same prompt encode once, which holds only if the wiring and the cache are both right. Reverting the call sites gives `2 failed, 62 passed`; restoring gives `64 passed`.

`make verify` passes locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved realtime prompt handling for faster, more consistent frame generation.
  * Added prompt caching to avoid redundant processing and improve reuse.
  * Improved support for short and long prompts across multiple model types.
  * Corrected embedding and pooled-output handling for enhanced compatibility.
  * Prevented duplicate prompt processing during realtime generation.
  * Fixed guidance handling for adapter and image-to-image generation.
  * Preserved both positive and negative prompt conditions during image generation.

### DIFF
--- a/worker/tests/test_engine.py
+++ b/worker/tests/test_engine.py
@@ -128,7 +128,9 @@ class _FakePipeline:
 
     def __call__(self, **kwargs):
         self.call_kwargs.append(kwargs)
-        return SimpleNamespace(images=[object()])
+        # The job paths encode the rendered image with encode_png, which needs
+        # a real PIL image, not the object() the frame paths can get away with.
+        return SimpleNamespace(images=[Image.new("RGB", (8, 8))])
 
     def encode_prompt(self, prompt, *, device, do_classifier_free_guidance,
                       num_images_per_prompt):
@@ -422,6 +424,77 @@ def test_realtime_frame_i2i_path_encodes_prompt():
     call_kwargs = pipeline.call_kwargs[0]
     assert "prompt" not in call_kwargs
     assert call_kwargs["prompt_embeds"].shape == (1, 2, 4)
+
+
+def test_generate_job_encodes_both_conditionings_at_positive_guidance():
+    """The text-to-image job call site must keep handing the pipeline both
+    conditionings. If it were switched to the realtime prompt helper without
+    a guidance argument, the helper's zero-guidance shortcut returns
+    positive-only embeddings, and diffusers 0.39.0 encodes an empty negative
+    rather than raising: every negative-prompted job would silently render
+    with the negative dropped. It is reachable on shipped defaults because
+    run_job applies manifest.with_defaults and sdxl-base declares guidance 6,
+    above diffusers' CFG threshold of 1.0.
+    """
+    engine = _fake_prompt_engine()
+    engine._pipelines = {}
+    pipeline = _fake_pipeline()
+    engine._pipeline = MagicMock(return_value=pipeline)
+    manifest = _clip_manifest()
+    loop = asyncio.new_event_loop()
+    try:
+        engine._generate(
+            manifest,
+            {"prompt": "w0 w1", "guidance": 6.0, "negative_prompt": "w3"},
+            lambda _: None,
+            loop,
+        )
+    finally:
+        loop.close()
+
+    call_kwargs = pipeline.call_kwargs[0]
+    if "negative_prompt" in call_kwargs:
+        assert call_kwargs["prompt"] == "w0 w1"
+        assert call_kwargs["negative_prompt"] == "w3"
+    else:
+        assert "negative_prompt_embeds" in call_kwargs
+
+
+def test_generate_i2i_job_encodes_both_conditionings_at_positive_guidance():
+    """The image-to-image job call site is the same regression surface:
+    ssd-1b declares guidance 7, so the pipeline must receive both
+    conditionings here too, never the realtime helper's positive-only
+    shortcut.
+    """
+    engine = _fake_prompt_engine()
+    engine._pipelines = {}
+    pipeline = _fake_pipeline()
+    engine._pipeline = MagicMock(return_value=pipeline)
+    manifest = Manifest(
+        id="vega-rt",
+        name="VegaRT",
+        capabilities=["text_to_image", "image_to_image", "realtime"],
+    )
+    buffer = io.BytesIO()
+    Image.new("RGB", (64, 48), (40, 80, 120)).save(buffer, "PNG")
+    loop = asyncio.new_event_loop()
+    try:
+        engine._generate_i2i(
+            manifest,
+            {"prompt": "w0 w1", "guidance": 7.0, "negative_prompt": "w3"},
+            lambda _: None,
+            loop,
+            buffer.getvalue(),
+        )
+    finally:
+        loop.close()
+
+    call_kwargs = pipeline.call_kwargs[0]
+    if "negative_prompt" in call_kwargs:
+        assert call_kwargs["prompt"] == "w0 w1"
+        assert call_kwargs["negative_prompt"] == "w3"
+    else:
+        assert "negative_prompt_embeds" in call_kwargs
 
 
 def test_sdxl_pooled_embedding_comes_from_first_chunk():
--- a/worker/tests/test_engine.py
+++ b/worker/tests/test_engine.py
@@ -113,18 +113,40 @@ def _fake_prompt_engine():
     return engine
 
 
+class _FakePipeline:
+    def __init__(self, *, dual=False):
+        self.tokenizer = _FakeTokenizer()
+        self.text_encoder = _FakeTextEncoder(4)
+        self.tokenizer_2 = None
+        self.text_encoder_2 = None
+        self.config = SimpleNamespace(force_zeros_for_empty_prompt=True)
+        self.encode_calls = 0
+        self.call_kwargs = []
+        if dual:
+            self.tokenizer_2 = _FakeTokenizer()
+            self.text_encoder_2 = _FakeTextEncoder(6)
+
+    def __call__(self, **kwargs):
+        self.call_kwargs.append(kwargs)
+        return SimpleNamespace(images=[object()])
+
+    def encode_prompt(self, prompt, *, device, do_classifier_free_guidance,
+                      num_images_per_prompt):
+        assert device == "cpu"
+        assert do_classifier_free_guidance is False
+        assert num_images_per_prompt == 1
+        self.encode_calls += 1
+        token_ids = self.tokenizer(
+            prompt, add_special_tokens=False, truncation=False)["input_ids"]
+        embeds = _FakeTensor((1, len(token_ids), 4))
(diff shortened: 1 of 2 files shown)


leonfullxr:
Two independent reviews found the same defect in this branch, and it was worse than the first description of it. Fixed in da4656d, recorded here because the way it happened matters more than the fix.

The commit that added the guidance guard also switched `_generate` and `_generate_i2i` to the realtime helper, which its brief did not ask for. Those job paths pass a settable guidance to the pipeline, and the helper was called without a guidance argument, so it took its `0.0` default, returned positive-only embeddings and cached them. diffusers 0.39 does not raise when `prompt_embeds` arrives without `negative_prompt_embeds`; it encodes an empty negative instead.

The second review established what makes that severe rather than theoretical: it is reachable on shipped defaults. `run_job` applies `manifest.with_defaults`, and `sdxl-base` declares guidance 6, `ssd-1b` 7 and `dreamshaper-lcm` 1.5, all above diffusers' classifier-free-guidance threshold of 1.0, so every batch text-to-image and image-to-image job with a negative prompt would have rendered with that negative silently dropped: the watermark, the blur, the text the user asked to exclude all stay in the image. `sd35-medium` escapes only because it takes the SD3 string path. It is deterministic, not a race, and it does not depend on whether the cache hits.

The fix is to put those two call sites back on `_prompt_kwargs`, which is where they were before this branch. The suggested alternative was to pass the real guidance so they route through the degrade path; that also works, and this way is smaller and removes the coupling instead of managing it. A job's prompt is new on every dispatch, so the cache was never going to help it.

What the suite says about all this: 151 tests passed with the regression present and 151 passed with it gone. No test drives either job path, so nothing could have caught it. That test is being added now.

The same review cleared the rest of the change, which is worth recording because it was checked rather than assumed:

- Two sessions sharing a pipeline cannot receive each other's conditioning. Every caller of the helper runs inside `async with self._gpu`, and `_run_to_completion` drains its thread before the lock unwinds, so the cache read, compare and write are serialised. A different prompt misses and re-encodes; an identical one returns the same tensors. The worst case is single-slot thrash between alternating sessions, which costs the speedup and never renders a wrong frame.
- Nothing frame-varying is cached. The entry holds `prompt_embeds` and, on SDXL, `pooled_prompt_embeds`; the canvas, seed, steps, strength and structure strength are all passed at the pipeline call.
- The memory is bounded and small: about 0.31 MiB per entry for SDXL at the 77-token window, one entry per pipeline, replaced rather than accumulated, and freed with the pipeline on eviction. A session that edits its prompt on every frame replaces the entry each time and cannot grow it.
- A cache miss cannot double encode. `_prompt_kwargs` runs first and returns before any encode for a short prompt; a chunked long prompt returns already carrying `prompt_embeds` and the helper returns it untouched; SD3 defers earlier still.

The one-entry cache is being kept as it is. When several sessions share one GPU, which is issue #294, two sessions with different prompts will thrash that entry and lose the saving, and that is the point at which it should become a small keyed cache rather than now.

leonfullxr:
Closing this in favour of #311, which implemented the same issue and merged first. That implementation is better than this one on both points that were open here, so this branch has nothing left worth landing.

Its cache belongs to the session rather than to the pipeline. A `PromptCache` is created by the `SessionRunner` and passed into `engine.frame`, so two sessions on one model cannot evict each other's entry, and the entry is dropped with the runner instead of needing a release path. The review comment on this PR asked for exactly that and it is what #311 already does.

Its cache is also safe on the job paths by construction. It reaches every prompt through the project's chunking encoder, which always returns both conditionings, so a job asking for guidance above the classifier-free threshold keeps its negative. This branch instead encoded without classifier-free guidance and needed a guard to stay correct, and a commit here then defeated that guard by routing `_generate` and `_generate_i2i` through the realtime helper with a defaulted zero guidance. Two reviews found it independently. Both approaches can be made correct; only one of them cannot be broken by a later edit that forgets the guard.

Three things checked before closing, so the decision rests on measurements rather than on preference.

Removing the `chunk_count == 1` early return did not change what jobs render, which was the one risk in #311 that nobody had verified. Encoding the same short prompt and negative both ways gives a maximum absolute difference of 0.000e+00 across `prompt_embeds`, `negative_prompt_embeds`, `pooled_prompt_embeds` and the negative pooled embedding.

The session cache engages on the shipped path and is worth about 14 ms a frame: 265.2 ms p50 on a hit against 279.2 ms with a fresh holder forcing every frame to encode, for `sdxl-turbo` at one step and 512 px. That number has to come from wall time around `engine.frame`, because `GeneratedFrame.gpu_ms` starts after `_prompt_kwargs` and would report no saving at all (issue #288).

The 21 to 24 ms measured on this branch is not comparable to that 14 ms, and neither is wrong. This branch was measured against the old path where diffusers encoded internally on every frame; #311's number is measured against re-running the project's chunker. Different baselines, same direction.

One cost of #311 is not recorded anywhere and is worth knowing: a single-encoder distilled job at guidance zero, `sd-turbo` for example, now pays one extra empty-negative CLIP forward per job, because the chunker always builds both embeddings where the early return used to skip it. Small against even a four-step denoise, and not a reason to restore the early return, which would disable the realtime cache for exactly the short prompts it exists for.

What survives from here is the pair of tests in #315: nothing on main drives `_generate` or `_generate_i2i` into the real `_prompt_kwargs` at a shipped guidance, so the hazard this branch demonstrated could return unnoticed.

PR #316, opened 2026-08-18
Decode realtime frames with a distilled VAE, in the path that runs

leonfullxr:
# Description

Reapplies the distilled preview decoder from the draft on `feat/issue-196-preview-vae` onto current main, and fixes the reason it would have shipped doing nothing. That draft is left untouched: adopt this branch, or cherry-pick its commits.

# Motivation

The draft had fallen 100 commits behind main, with conflicts in `engine.py`, `test_engine.py`, `test_manifests.py`, `manifests.py` and two documents. More importantly it predates the T2I-Adapter conditioning change. `_frame` now splits on `manifest.t2i_adapter`: the adapter branch renders and returns before reaching `_preview_decoder`, which sits in the image-to-image branch below it. Every shipped realtime model declares an adapter, `vega-rt` and `sdxl-turbo` both, so a realtime frame never reached the decoder.

Measured on the branch before the fix, `vega-rt` rendered at 335.2 ms p95 with the decoder configured and 340.5 ms with it removed. The same number, because the decoder never loaded, while 154 tests passed in both states.

# Functionality

Both branches of `_frame` now share one render helper: the pipeline returns latents, `AutoencoderTiny` decodes them without dividing by the VAE scaling factor, and the image processor postprocesses the result. On any exception the decoder is deferred for 60 seconds and the frame is rendered again through the full VAE, so a bad or missing decoder costs a slower frame rather than a lost one.

The load stays lazy, on the frame path. Warming it during `load_model` was tried and reverted: it built the realtime pipeline early, costing 0.15 GiB and a pipeline that a queued job on the same model never uses, and `vega-rt` serves jobs too. A `snapshot_download` prefetch was tried and reverted as well, because it put a network round trip on every model load and could not be limited to a cold cache: a repo snapshot is never complete when `from_pretrained` fetches only the files it needs, so a local-only probe always misses. What makes lazy acceptable is that `warmup_realtime` calibrates by rendering frames before `hello`, so the default realtime model's decoder is built before the worker registers and before any session exists. A realtime model that was not the warmed default, and the retry after a deferral, can still pay a cold fetch under the GPU lock.

# Details

Measured on the reference RX 7600 XT, `vega-rt` at its four steps and 512 px, 20 timed frames through `engine.frame` after a discarded warmup frame, and confirmed with the two cases run in both orders:

| Decoder | p50 | p95 | Slots at a cap of 4 |
|---|---|---|---|
| Full VAE | 331.9 ms | 336.4 ms | 1 |
| `madebyollin/taesdxl` | 216.5 ms | 219.8 ms | 2 |

Slots come from `floor(500 / p95)`, so this is what admits a second concurrent session, and it is the prerequisite for any end-to-end multi-user test: until now a worker advertised exactly one slot and a second session could not be opened at all.

Quality was accepted separately, by decoding one denoised latent both ways so nothing but the decoder differed: mean absolute difference 10 of 255 across four subjects at 22 to 28 dB, with the distilled decoder retaining 108 percent of the full decode's local gradient. The image a user keeps is still a queued job through the full VAE.

Three review findings from this branch are worth reading in the diff rather than only here. Clearing the calibrated slot count on a deferral does not reach a live worker, because `realtime_slots` crosses the wire in `hello` alone and heartbeats carry the observed p95 and `slots_in_use` but not capacity; the comment now says so, and closing that gap needs slots on the heartbeat or the worker-side refusal issue #270 still wants, which is not this PR. A frame whose decode raises reuses a generator that has already advanced, so that one image is not what a clean full-VAE render at the same seed would give; the next frame re-seeds. And a test pins the 60-second promise on the adapter path: after a decode failure the next frame makes exactly one pipeline call with no `output_type`, so the double denoise happens on the failing frame only.

Two earlier revisions of this PR quoted load-time costs of about three seconds for the warming and prefetch attempts. Those were measurement artifacts: the first manifest in that script paid one-time warmup, and swapping the order moved the three seconds to the other row. The pipeline and VRAM difference was structural and stands; the timing did not and has been removed from the code comment as well.

`make verify` passes locally, 158 worker tests.

## Summary by CodeRabbit

- **New Features**
  - Added an optimized decoder for faster real-time image previews.
  - Improved preview performance for text-to-image, image-to-image, and adapter-based workflows.
  - Preview processing now supports increased calibrated capacity.

- **Bug Fixes**
  - Previews automatically fall back to the standard decoder when optimized decoding fails.
  - Added automatic retry behavior after temporary decoder failures.
  - Improved reliability during image post-processing and WebP encoding.


leonfullxr:
Two of the three findings are fixed in b6b4862; the third is deliberately out of scope and now has its own issue.

**Slot arithmetic.** Correct, and the error came from the draft this branch reapplies. `floor(500 / 184.9)` is two, not three, and a third slot needs a frame under 166.7 ms. Rather than fix the one number by eye I checked every figure in that entry against `slots_from_frame_ms`: 184.9 gives 2, the measured 219.8 gives 2, 336.4 gives 1, and the encoder-swap paragraph's 94.0 ms really is 5. The entry now also carries what the shipped conditioned path measures, 216.5 ms p50 and 219.8 p95 against 331.9 and 336.4 on the full VAE, which is the same one-slot-to-two move at the step count the manifest ships.

**The attribute constant.** Done. Worth recording why, because my first comment on it had the reasoning backwards: keying the test by `_PREVIEW_DECODER_ATTR` does not catch a rename, it makes the test survive one. What it buys is that the assertion is about behaviour, that eviction removes whatever attribute the engine actually stores the decoder under, rather than about a spelling that would fail for the wrong reason after a rename.

**Invalidating advertised capacity after a decoder failure.** Real, and not fixable inside this PR. The engine already clears its calibrated slot count on a deferral; the problem is that the clear cannot reach the API, because `realtime_slots` crosses the wire only in `hello` and heartbeats carry the observed p95 and `slots_in_use` but not capacity. So a worker that calibrated two sessions and then falls back keeps both, each frame now costing full-VAE time, and two serialised sessions are about 670 ms apiece.

Closing that needs either capacity on the heartbeat or a live `session_refused` carrying its `control_generation`, which is the same missing message issue #270's remaining half wants, and the session design merged in #305 is explicit that a worker cannot end a session on its own. Both are protocol changes that want designing together rather than one landing here and being widened later, so this is issue #317, and the code comment at that site says what the clear does and does not achieve instead of implying the gap is closed.

leonfullxr:
On the two things the merge-risk note names.

**The mutable remote dependency is pre-existing practice, not something this adds.** No model reference in the repository pins a revision, and there is no revision plumbing to pin one with: ten manifests name a bare `source`, four add a `vae`, three a `lora`, two a `t2i_adapter`. `madebyollin/sdxl-vae-fp16-fix` is already an unpinned dependency of four shipped models, from the same author as `taesdxl`. So this branch adds one more instance of an existing pattern rather than a new class of risk, and pinning is a manifest schema change plus engine plumbing plus a decision about whether shipped manifests are pinned at release. Filed as #319 with the three consequences separated, reproducibility, recorded measurements, and supply chain, rather than bolted onto this PR.

**The capacity mismatch is explicitly accepted, with the reason in the code.** The engine clears its calibrated slot count on a deferral; the clear cannot reach the API, because `realtime_slots` crosses the wire only in `hello`. A worker that calibrated two sessions and then falls back keeps both at full-VAE speed, about 670 ms serialised, and previews keep arriving so nothing looks broken. Issue #317 covers it, and the comment at that site says what the clear does and does not achieve rather than implying it is handled. It needs capacity on the heartbeat or a live `session_refused` carrying its `control_generation`, which is the same missing message #270's remaining half wants, so it wants designing with that rather than here.

Worth being precise about what this branch is and is not verified to do. It is measured, under both orderings, to take `vega-rt` from 331.9 ms p50 and 336.4 p95 to 216.5 and 219.8, so calibration advertises two slots where it advertised one. Its quality was accepted by decoding one latent both ways. What has never been tested is two sessions actually rendering at once end to end, because until this branch a worker advertised exactly one slot and a second session could not be opened at all. That test becomes possible here and has not been run.

PR #318, opened 2026-08-19
Collect the blobs of a job that failed without a verdict

### LINKED ISSUE #312: A job that fails without a verdict never has its blobs collected
# Description

A job can reach a terminal state without any worker verdict, and nothing then attempts to delete what its attempts uploaded. `sweep_stalled_jobs`, `on_worker_lost` and `recover` all funnel into `requeue_or_fail`, whose failing branch calls `mark_failed` and nothing else. The two paths that call `purge_attempt_blobs` are the worker verdicts, so neither runs.

# Motivation

Concretely: attempt 2 uploads its PNG, the worker dies before sending `job_done`, the sweeper fails the row past its one retry. The object is on disk or in the bucket, no asset row names it, and no sweep looks for it. It is the same silent leak #254 exists to remove, arriving through the door #254 did not cover.

It predates the pending-deletes work: that branch retries deletes that were attempted and failed, and this is a delete that is never attempted.

# Functionality

Have the no-verdict terminal paths collect the same keys the verdict paths do. `requeue_or_fail` knows the job id and can read the user id and attempt from the row, which is everything `storage_keys_for_attempt` needs, and a failure there records into `pending_deletes` like any other.

Worth checking whether the requeue branch should do it too: a requeue to attempt 2 leaves attempt 1 objects behind, and today the success path collects those only if the job later succeeds.

# Details

`backend/app/jobs.py`, `requeue_or_fail` and its three callers. `purge_attempt_blobs` already walks every attempt up to the one it is given, so it is the right function to reuse.

Found by kimi-k3 while reviewing PR #309, which closes #254.

leonfullxr:
Closes #312. Implemented by OpenCode (deepseek-v4-flash, max effort) to a design I fixed in the brief; gated and mutation-checked by me.

A job can reach a terminal state with no worker verdict: the stall sweeper past its one retry, a worker that dies, a restart that recovers a running row. All three funnel into requeue_or_fail, whose failing branch called mark_failed and returned, while the only two callers of purge_attempt_blobs are the worker verdicts. Attempt two uploads, the worker dies before job_done, the sweeper fails the row, and the object sits there with no asset row naming it: the same silent leak #254 removes, through a door it did not cover.

The failing branch collects now, and only when mark_failed reports that it committed. A refusal means the row went terminal or moved to a later attempt while this failure was in flight, and the keys then belong to whoever won.

## The requeue branch is deliberately untouched

It leaves attempt one objects behind, which the success path collects only if the job later succeeds. Collecting at requeue time is not safe on S3: the presigned PUT of the attempt being replaced stays valid for its whole TTL and needs no backend check, so a stale worker can land bytes after the purge, and the If-None-Match guard then keeps them with no collector left. The local backend re-checks against the inflight entry and would be safe. That asymmetry is storage-specific and wants its own issue rather than a guess here. The delegate surfaced this while implementing, and a test pins the current behaviour so a future change to it is deliberate.

## Verified

Five tests, each proved by breaking what it covers. One of them took three attempts to make honest: my first version asserted the refusal guard against a job at attempt one, where requeue_or_fail requeues and never reaches the guard at all, so it passed against a mutation that removed the guard entirely. It now drives the job past its retry first.

make verify green: 329 backend, 141 worker.

## Summary by CodeRabbit

* **Bug Fixes**
  * Failed jobs now reliably reach their final failed state even when upload cleanup encounters an error.
  * Uploads from failed and earlier attempts are cleaned up appropriately.
  * Uploads belonging to replacement attempts are preserved during failure handling.
* **Tests**
  * Added coverage for stalled, disconnected, retried, and cleanup-failure scenarios.

### DIFF
--- a/backend/app/jobs.py
+++ b/backend/app/jobs.py
@@ -1677,7 +1677,14 @@ async def requeue_or_fail(job_id: uuid.UUID, reason: str) -> None:
             publish(job_id, {"state": "queued", "attempt": 2})
             logger.info("job %s requeued after %s", job_id, reason)
             return
-    await mark_failed(job_id, reason)
+        user_id = job.user_id
+        attempt = job.attempt
+    if await mark_failed(job_id, reason):
+        # A verdictless failure can still have uploaded: nothing downstream
+        # names those objects and no verdict path ran, so this is their only
+        # collector. The commit is durable before any delete, and a failing
+        # delete must not fail the job: the purge records it instead.
+        await purge_attempt_blobs(user_id, job_id, attempt)
 
 
 def on_worker_lost(worker: realtime.Worker) -> None:
--- a/backend/tests/test_jobs.py
+++ b/backend/tests/test_jobs.py
@@ -860,6 +860,169 @@ def test_a_retry_does_not_leave_the_earlier_attempt_behind(monkeypatch):
 
 
 @pytest.mark.db
+def test_a_stalled_job_failed_past_its_retry_collects_its_uploads(monkeypatch):
+    """A job failed by the sweeper past its one retry never got a worker
+    verdict, so requeue_or_fail's failing branch is the only collector its
+    uploads have (issue #312)."""
+    monkeypatch.setenv("JOB_STALL_SECONDS", "600")
+    from app.settings import get_settings
+    get_settings.cache_clear()
+    try:
+        with TestClient(app) as client:
+            with client.websocket_connect("/api/v1/fleet") as worker:
+                fleet_hello(worker, "w-stall-fail")
+                job_id = client.post(
+                    "/api/v1/generations",
+                    json={"model_id": "sd-test", "params": {"prompt": "stall fail"}},
+                ).json()["job_id"]
+                first = worker.receive_json()
+                assert first["job_id"] == job_id
+
+                force_one_requeue(client, job_id)
+                second = worker.receive_json()
+                assert second["job_id"] == job_id
+                # Attempt 2 uploads its PNG and then dies without a verdict.
+                second_key = urlsplit(second["upload"]["url"]).path.rsplit(
+                    "/api/v1/files/", 1)[-1]
+                assert put_upload(client, second["upload"],
+                                  png_bytes()).status_code == 200
+                storage = jobs.get_storage()
+                assert storage.path(second_key).exists()
+
+                # Stall again: past the one retry, the sweeper fails the row.
+                jobs.last_progress_at[uuid.UUID(job_id)] = 0.0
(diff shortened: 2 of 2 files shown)


leonfullxr:
Review by `opencode-go/kimi-k3`: **ship it**, no findings. It did correct my reasoning, which is worth recording because the conclusion survived and the argument did not.

**What I claimed:** collecting attempt one's objects in the requeue branch is unsafe, because on S3 a stale worker can land bytes after the purge and the `If-None-Match` guard then keeps them with no collector left.

**What is actually true:** every terminal path of attempt two already re-collects attempt one. The success path deletes `range(1, attempt)`, the reported-failure path purges 1..2, and this diff makes the verdictless path purge 1..2. So a late upload would still be swept when the job ends, and "no collector left" was wrong. The conclusion stands for a smaller reason: collecting at requeue time buys reclaiming storage sooner in exchange for a purge on the hot retry path, which wants measuring rather than guessing. I have amended the commit message rather than leaving the wrong argument in the history, and I am **not** filing the asymmetry issue I said I would, because its premise was incorrect.

The genuinely uncollectable window is narrower and not specific to requeues: on S3 a presigned PUT landing after the last purge naming its key recreates it, bounded by the one-hour URL lifetime. That is the leak #278 already describes, so it needs no new issue.

Its answers to the four questions, none of which I need to re-derive:

- **Cannot delete an owned object.** The failing branch is reachable only at `running` and `attempt == 2`; `attempt` has exactly one write site, `user_id` is set at creation, and rows are never deleted. Once `mark_failed` commits, the row is terminal forever, so keys 1..2 can never be re-owned.
- **The two-session read cannot go stale in a way that matters**, and for a stronger reason than "attempt only grows": a running row at attempt 2 can only go terminal, since the requeue branch requires `attempt == 1` and moves through `queued` under the same lock. Between the first lock and `mark_failed`'s lock the row is either unchanged or terminal, and a terminal row means refusal and no purge.
- **The ordering matches the durability rule**: `mark_failed` commits in its own session before the purge, and per-key failures are recorded rather than raised.
- **All five tests are honest**, including the refusal one after I rewrote it; it flagged one nitpick, that the refusal test lacks the `JOB_STALL_SECONDS=600` guard its siblings carry, though the patched `mark_failed` neutralises the sweeper anyway.

One optional hardening it offered and I did not take: have `mark_failed` return the committed row's `(user_id, attempt)` from its own locked read, so the purge keys come from the same lock as the commit. Today's safety rests on the branch structure rather than on anything enforced, so that would make it structural. Worth doing if that function is next touched.

`make verify` green: 329 backend, 141 worker.

PR #359, opened 2026-08-21
Add readiness, a bounded database pool, and the accounts upgrade guard

leonfullxr:
# Description

Add a readiness probe, a bounded database connection pool, and a one-way guard that stops an installation with accounts enabled from restarting without them.

# Motivation

The load balancer had only a liveness route, which answers from process state and cannot tell whether PostgreSQL and the asset store are usable. The database engine used no pool, so every request opened a new connection. An installation also had no persistent record of the authentication mode it had already adopted, so a configuration change could silently return it to unauthenticated operation.

# Functionality

`GET /api/v1/ready` checks PostgreSQL and the configured asset store and returns 200 or 503 with an empty body. The runtime engine uses a bounded pool. A new `installation_auth_state` table records that an installation enabled accounts; startup refuses to continue if the configured mode contradicts that record. A dependency guard rejects account-dependent requests with 503 while the database is unavailable, and a PostgreSQL advisory lock keeps two account startups from running at the same time.

# Details

Base is `security/auth-r1-endpoint-hardening` (PR #358); it retargets to `main` after that merges.

The mode literal is now `none` or `accounts`, and `auth_methods` reports `password`, `google`, and `github`. Unknown OAuth provider names are dropped instead of forwarded. `accounts` still refuses to boot because no account route exists yet, so the guard and the advisory lock are inert rails for the rounds that add sign-up and sign-in.

The database test harness now keeps setup, routes, and disposal on one event loop, which the bounded pool requires. Local verification passed 391 backend tests, 194 worker tests, the frontend suite, Ruff, mypy, Svelte checks, the production build, and CSP verification.

## Summary by CodeRabbit

- **New Features**
  - Added a readiness endpoint that reports database and asset-storage availability.
  - Added persistent account-mode tracking and coordinated startup handling.
  - Limited authentication modes to `none` and `accounts`, with password and supported OAuth providers.

- **Bug Fixes**
  - Improved startup validation and prevented inconsistent authentication configuration.
  - Added bounded readiness checks for local and S3 storage.

- **Documentation**
  - Documented readiness responses and local development behavior.

### DIFF
--- a/backend/app/db.py
+++ b/backend/app/db.py
@@ -17,7 +17,7 @@ from pathlib import Path
 from alembic import command
 from alembic.config import Config
 from fastapi import HTTPException
-from sqlalchemy import select
+from sqlalchemy import select, text
 from sqlalchemy.ext.asyncio import (
     AsyncEngine,
     AsyncSession,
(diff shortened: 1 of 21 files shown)


leonfullxr:
@coderabbitai review

PR #360, opened 2026-08-21
Add the account schema and the root key ring

leonfullxr:
# Description

Add the account schema and the versioned root key ring the accounts release is built on.

# Motivation

Everything the accounts release does needs a place to put it and a way to keep its secrets. Without a schema there is nowhere to record an identity, a session, a capability or a queued email, and without a key ring every stored secret would sit under one static environment variable with no way to rotate it and one blast radius for every subsystem.

# Functionality

`ROOT_KEYS` holds comma separated `version:base64key` entries, newest first. The first entry is the active write key and every entry stays readable, so rotation is active write, multi read, re-encrypt, then remove, with no downtime. Purpose keys come from HKDF-SHA256 with the purpose in the info string. Values are sealed with AES-GCM and carry their key version, and the purpose plus any row binding go into the associated data. Every refusal raises: a key that has left the ring loses its plaintext rather than falling back to another key. The ring is separate key material from `FLEET_TOKEN_KEY`.

The schema adds `auth_identities`, `sessions`, `auth_tokens`, `invitations`, `auth_factors`, `recovery_codes` and `mail_outbox`, and gives `users` a `state` and a `mail_verified`. PostgreSQL holds every rule rather than the application: one password identity per account, one provider subject per account, a provider identity that can never carry a local credential, one open invitation per address, one TOTP factor per account, and role and state constrained to the designed values. Deleting an account takes its identities, sessions, tokens, factors and recovery codes with it.

`installation_auth_state` now records the root key version the installation writes with, and startup refuses when the ring no longer holds it.

# Details

Base is `feat/auth-r2-foundation-guard` (PR #359); it retargets upward as the stack merges.

This round is schema and key material only. There are no routes, no login flow, no password hashing and no TOTP verification; those belong to the rounds that add them, and `AUTH_MODE=accounts` still refuses to boot.

Accounts are unique on `lower(btrim(email))` through a unique functional index, and invitations use the same expression in a partial index, so there is no stored normalized column to fall out of step with the address it was derived from.

Deleting an administrator does not revoke the invitations they minted, because an invitation is open while `accepted_at` and `revoked_at` are both null and deleting the inviter sets neither. That is a cross-row rule no constraint can express, so it belongs to the round that implements account deletion; it is recorded here because the `SET NULL` on `invited_by` is what makes it easy to miss later.

`cryptography` is a new dependency and supplies both HKDF and AES-GCM. Nothing here is hand rolled.

Migration 0015 was checked to round trip to 0014 and back, and its schema was compared against the ORM metadata with no difference. Local verification passed 447 backend tests, 194 worker tests, the frontend suite, Ruff, mypy, Svelte checks, the production build and CSP verification.

## Summary by CodeRabbit

- **New Features**
  - Added secure, versioned encryption for account secrets, including key rotation and re-encryption.
  - Added authentication data storage for identities, sessions, tokens, invitations, security factors, recovery codes, and email delivery.
  - Added account verification, account-state tracking, normalized email uniqueness, and validation rules.
  - Added startup checks to detect missing or unusable encryption keys.

- **Documentation**
  - Documented the new `ROOT_KEYS` configuration and encryption design decisions.

### DIFF
--- a/backend/app/db.py
+++ b/backend/app/db.py
@@ -27,6 +27,7 @@ from sqlalchemy.ext.asyncio import (
 )
 from sqlalchemy.pool import NullPool
 
+from app.keyring import KeyRingError, get_key_ring
 from app.settings import get_settings
 from app.tables import InstallationAuthState, User
 
@@ -89,6 +90,7 @@ async def connect() -> bool:
     engine = create_async_engine(async_url(settings.database_url), pool_size=5, max_overflow=10)
     session_factory = async_sessionmaker(engine, expire_on_commit=False)
     await validate_startup_auth_mode(settings.auth_mode)
+    await validate_startup_key_ring()
     local_user_id = await _ensure_local_user(session_factory)
     return True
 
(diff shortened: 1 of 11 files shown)


leonfullxr:
@coderabbitai review

PR #403, opened 2026-08-24
Export, deletion, restore, and purge

### LINKED ISSUE #402: Export, deletion, restore, and purge
# Description

Let an account take everything this install holds about it, ask to be deleted, be restored inside a window, and be purged after it.

# Motivation

R13 gave an account a state that can stop it. Nothing could take the data out, ask to leave, or finish the job, so `deletion_pending` had no way in and no way out.

# Functionality

- `GET /api/v1/account/export`: the account, its identities and its generations with their assets, paged out of PostgreSQL and streamed, carrying no secret of any kind.
- `DELETE /api/v1/account`: stops the account now, destroys nothing for 30 days, and records the state to come back to.
- `POST /api/v1/users/{id}/restore`: admin only, idempotent, one level of prior state.
- A sweep that purges the objects, then the rows that named them, then the user row.

# Details

The purge order is what the foreign keys demand: an asset points at the job that made it and a job at the account. One account per transaction, so a pass that dies halfway loses nothing. The last administrator may delete their own account, because an install with nobody in charge can be recovered offline and a person cannot be untrapped.

leonfullxr:
Closes #402

# Description

An account can take everything this install holds about it, ask to be deleted, be restored inside a window, and be purged after it.

# Motivation

R13 gave an account a state that can stop it. Nothing could take the data out, ask to leave, or finish the job, so `deletion_pending` had no way in and no way out.

# Functionality

`GET /api/v1/account/export` returns the account, its identities and its generations with their assets, as one JSON document paged out of PostgreSQL and written as it goes. A library of ten thousand generations never has to fit in memory, here or in the process that asked for it. It carries no password hash, no session hash, no TOTP secret and no recovery code, because a file that leaves the building takes whatever is in it wherever it goes.

`DELETE /api/v1/account` stops the account immediately and destroys nothing for 30 days. Sessions are revoked, outstanding links spent, realtime sockets closed, queued and running jobs cancelled. Stopping is what the person asked for and destroying is what they cannot undo; those do not have to happen in the same moment.

`POST /api/v1/users/{id}/restore` puts an account back where it was, one level deep, and is idempotent. An account that was suspended when it asked to be deleted comes back suspended: a restore undoes the deletion, not everything before it. An account that was never waiting to be deleted answers 409.

A sweep purges what is past the window: the objects first, since the asset row is the only thing that names them, then the assets, then the jobs, then the user row. One account per transaction, so a pass that dies halfway loses nothing and the next finds the rest where it was.

# Details

The last administrator may delete their own account. An install with nobody in charge can be recovered offline, and an administrator held hostage by their own install cannot.

Audit rows carry plain ids with no foreign key, so what an administrator did survives the account they did it to.

Migration 0021 adds `users.prior_state` and `users.deletion_requested_at`. Round-trip checked (upgrade, downgrade to 0020, upgrade); ORM drift unchanged at the 12 pre-existing entries. 13 new tests. `make verify` passes: 826 backend, 246 worker, frontend build and CSP check, 38 Mermaid diagrams.

## Summary by CodeRabbit

- **New Features**
  - Added account data export as streamed JSON, excluding credentials and secrets.
  - Added account deletion with immediate access revocation and a 30-day restoration window.
  - Added administrator restoration for accounts pending deletion.
  - Added automatic cleanup after the restoration period, including stored assets and related data.
  - Active work is cancelled when deletion is requested.

- **Documentation**
  - Documented account export, deletion, restoration, retention, and cleanup behavior.

- **Bug Fixes**
  - Improved cleanup resilience by recording failed asset removals for retry.


leonfullxr:
@coderabbitai review

PR #406, opened 2026-08-24
Close the gap between reclaiming and claiming an install

leonfullxr:
# Description

CodeRabbit posted a second review round on #405 after that PR was merged. This is the rest of it.

# Motivation

I merged #405 while its review was still running and did not read what came back. Five findings, all valid.

# Functionality

- The offline reclaim checked that nobody held an identity, closed that transaction, and minted a setup link in another. A claim landing in the gap left an operator holding a link refused the moment they spend it. Both halves are one transaction now, behind an advisory lock that `POST /api/v1/auth/setup` takes as well.
- The cloud guide said `collapse` is not a cloud command, which reads as a guard. It is a policy: anybody who can open an ECS Exec session in the API task can run it. The guide says that outright now, and says what to do if it is not tolerable.
- ECS Exec from a private subnet needs NAT egress on 443 or an `ssmmessages` interface endpoint, and its permissions belong to the task role, not the execution role. Neither was documented, and both fail as a hanging command rather than an error.
- The recovery procedure turned session logging off and Exec on, and never put either back. The next Exec session anybody opens for an unrelated reason would then go unrecorded, with the shell still available.

# Details

The lock is the interesting half. `mint_setup_token` now takes an optional session so the check and the mint share one transaction; `claim()` takes the same lock, so the reclaim's check means something.

Two new tests, both neuter-checked: one that the reclaim is atomic, and one that the claim route actually waits on the lock, which is the half a test could easily fake by taking the lock itself.

`make verify` backend passes: 858 tests.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved coordination between setup-token creation, installation claims, and recovery actions to prevent conflicting concurrent operations.
  - Setup claims now wait for recovery actions to complete, reducing invalid or stale setup links.
  - Setup-token creation and recovery are refused after an installation has been claimed.
  - Recovery operations now verify account state before issuing new setup links.

- **Documentation**
  - Expanded AWS recovery guidance, including ECS Exec requirements, networking, logging, and credential-retention warnings.
  - Clarified when installation setup claims are refused and how recovery options differ.

### DIFF
--- a/backend/app/enable.py
+++ b/backend/app/enable.py
@@ -56,6 +56,11 @@ async def mint_setup_token(session: AsyncSession | None = None) -> str:
         raise RuntimeError("database unavailable")
     async with db.session_factory() as owned:
         async with owned.begin():
+            # The same lock the caller-supplied branch is already inside, so
+            # every way of minting a setup link is serialized with claiming
+            # one however it was reached.
+            await owned.execute(text("SELECT pg_advisory_xact_lock(:key)"),
+                                {"key": SETUP_LOCK})
             await _replace_setup_token(owned, token)
     return token
 
--- a/backend/app/enable.py
+++ b/backend/app/enable.py
@@ -13,8 +13,10 @@ from datetime import datetime, timedelta, timezone
 from fastapi import APIRouter, Depends, HTTPException
 from pydantic import BaseModel
 from anyio import to_thread
-from sqlalchemy import delete, func, select, update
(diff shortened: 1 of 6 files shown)


leonfullxr:
@coderabbitai review

leonfullxr:
@coderabbitai review

leonfullxr:
@coderabbitai full review

PR #426, opened 2026-08-25
Let an account take its second factor off, and the machine clear it

### LINKED ISSUE #419: A lost authenticator is unrecoverable: nothing removes a second factor, including the offline commands
# Description

Once an account enrols TOTP, nothing can un-enrol it. `app/factors.py` exposes three routes - answer a challenge, start an enrolment, confirm an enrolment - and no route that turns the factor off. The only `delete(AuthFactor)` in the codebase is inside `confirm_enrolment`, where a new authenticator replaces the previous one, so replacing a factor is possible and removing one is not.

The gap does not stop at HTTP. Neither offline path clears it either:

- `python -m app.recovery EMAIL` sets a password and never touches `auth_factors`.
- `operator.py` reads `AuthFactor` only to re-encrypt it during key rotation. `reclaim --restore` leaves it in place.

So the sequence is: enrol TOTP, lose the authenticator, spend or lose the recovery codes, and the account is locked out with no way back. The offline administrator recovery command does not help, because it hands back a password link and the TOTP gate still stands between that password and a session.

# Motivation

For an ordinary account this is a support problem an administrator cannot solve: there is no administrator action that clears a factor either.

For the last administrator it is worse. "Every way back into a locked install is a command at the machine" is a recorded decision, and the whole point of the offline recovery command is that an install is never unrecoverable. That promise does not currently hold for an administrator who enrolled TOTP and lost both the authenticator and the codes: the command runs, prints a link, sets a password, and the next sign-in stops at a challenge nothing can answer. The only remaining route is `make auth-collapse`, which destroys every account on the installation to get back in.

It is also the reason the second factor has to stay optional. That is worth making a decision rather than an accident.

# Functionality

Decide which of these is intended, then make the code and the documents agree:

1. An account may remove its own factor, with recent authentication, the way it can change a password. Sessions rotate on the change like every other credential event.
2. An administrator may clear another account's factor, audited with the target, subject to the last-administrator guard.
3. The offline recovery command clears the factor when it sets the password, so the machine remains the way back it is documented to be.
4. It stays impossible on purpose, in which case that belongs in `docs/decisions.md` with its reasoning, and the recovery documentation in `docs/self-hosting.md` and `docs/aws-setup.md` needs to say plainly that recovery does not clear a second factor.

Whichever is chosen, `POST /api/v1/account/totp` starting an enrolment while one already exists is the closest thing to a removal today and is worth reading alongside it.

# Details

- `backend/app/factors.py`: routes at `/api/v1/auth/totp`, `/api/v1/account/totp`, `/api/v1/account/totp/confirm`. The `delete(AuthFactor)` is inside the confirm path.
- `backend/app/recovery.py`: no reference to `factors` or `AuthFactor`.
- `backend/app/operator.py`: `AuthFactor` appears only in the rotation sweep.

Found while writing the shipped-TOTP decision entry for #413. It was deliberately left out of that entry rather than having a rationale invented for it, since whether this is a decision or an omission is exactly the open question.

leonfullxr:
# Description

An account can now remove its second factor, at the same cost as adding one, and `make auth-clear-factor` handles the account that has lost every way to prove it. Closes #419.

# Motivation

Nothing removed a factor. Three routes existed, none of them turned it off, and the only `delete(AuthFactor)` in the codebase sits inside confirming a replacement. So "optional for every role", which the decision entry says and the code otherwise honours, was true of turning a factor on and not of turning it off. An account whose authenticator broke could only spend recovery codes until they ran out, and then it was unreachable.

For the last administrator it was worse than an inconvenience. `python -m app.recovery` prints a password link and never touches `auth_factors`, and `operator.py` reads them only to re-encrypt during key rotation, so the challenge still stood behind the new password. An install could reach a state recoverable only by `auth-collapse`, which destroys every account on it. "Every way back into a locked install is a command at the machine" is a recorded decision, and it was not true.

# Functionality

- `DELETE /api/v1/account/totp` takes recent authentication and a code, either a current one from the authenticator or a recovery code. The factor and its recovery codes go together.
- `make auth-clear-factor EMAIL=...` removes a factor from the machine, for the account that can present neither.
- A decision entry recording the bar and why a session alone does not meet it.

# Details

**Why removal costs a code.** `start_enrolment` deliberately writes nothing until a code answers, and its docstring says why: writing first "made this endpoint the cheapest way to turn the second factor off: one request from a stolen session, no code, and no notice to anybody". Removal is that same act from the other side. A route that took only a session and recent authentication would reintroduce exactly the hole enrolment is written to avoid, and it would be the weakest point in the whole second-factor design.

A recovery code counts, which is the case that otherwise ends badly. Somebody whose phone died still holds nine ways to prove the factor is theirs, and spending one to turn it off is better than spending them one at a time on logins until the account is unreachable.

**Why the offline command exists.** An account that has lost the authenticator and every code has nothing to present, and no route can safely help it: a route that clears a factor without asking for one is precisely the route worth stealing a session for. So it is a command at the machine, like every other way back in. It removes the factor and its codes and touches nothing else; the account keeps its password, its role and its sessions, and can enrol again whenever it likes.

**Inherited protections, checked rather than assumed.** `DELETE` is not in `SAFE_METHODS`, so the route gets the exact-Origin and CSRF-header check like every other mutation. The router already carries `require_accounts_mode`, so it is a `404` in `none` mode. A suspended account is refused every non-safe method, so it cannot clear its factor while read-only.

**Three guards, three neuter checks.** Drop the code check and `removing_the_factor_needs_the_factor` fails. Drop the recent-authentication check and `removing_needs_recent_authentication` fails. Leave the recovery codes behind and two fail, the HTTP one and the offline one, because both assert the codes leave with the factor.

**Rejected alternatives are in the decision entry**, including the two that look reasonable until written out: letting a password reset clear the factor, where control of an inbox would be enough and mail is what the second factor exists to survive, and an administrator route to clear somebody else's, which is the same route worth stealing a session for one privilege level up.

`make verify` from the worktree with `PYTHONPATH` forced to it: 904 backend, 247 worker, frontend and CSP green.

**This one was a judgement call rather than a bug fix**, and it is easy to revert if you would rather have the other answer: leaving a factor unremovable on purpose and documenting the lockout in the self-hosting and AWS guides. I took this one because both halves follow from decisions already recorded rather than making new policy, and because the alternative leaves an account's own security setting as a trap.

## Summary by CodeRabbit

- **New Features**
  - Added the ability to remove a TOTP authenticator using a valid TOTP or recovery code.
  - Added an administrative fallback for clearing an account’s second factor when authentication codes are unavailable.
  - Passwords, roles, sessions, and other account data remain unchanged.

- **Bug Fixes**
  - Removing a second factor now also removes its associated recovery codes.

- **Documentation**
  - Documented the removal endpoint, authentication requirements, and administrative recovery procedure.


leonfullxr:
Moved to draft. The review found five P1s and one of them takes the ground out from under the design, so this needs reworking rather than patching.

**The command does not work at all.** `_run_clear_factor` calls `_connected(clear_factor(email))`, but `_connected` is async and expects a callable, and nothing runs it. So `make auth-clear-factor` builds a coroutine, never awaits it, and prints success while the factor stays exactly where it was. The test called `operator.clear_factor` directly and so never exercised the command it was supposed to prove. Testing the helper instead of the thing that ships is the whole of that mistake.

**The justification is false.** This PR argued that removal should cost a code because enrolment already does. It does not. `confirm_enrolment` verifies the submitted code against the **new** secret from `start_enrolment` and then deletes whatever factor the account had. Nothing asks for a code from the existing factor, so a session with recent authentication can already replace somebody's second factor with one it controls, in two requests, and the account finds out when its authenticator stops working. Filed as #427, and it is more urgent than the gap this PR was opened to close.

That inverts the order of work. Removal cannot be justified by parity with enrolment until enrolment has a bar, so #427 comes first and this follows it.

**Also found, and all real:**

- A suspended account can remove its factor. The suspended check lives in `current_user`, and this route depends on `current_principal`, which does not carry it. I checked this before opening the PR, read the wrong dependency, and told myself it was covered.
- Removal revokes no sessions and rotates nothing, and the offline path deliberately keeps every session. A stolen session survives the security change that was meant to end it, including a copy of the very session that performed the removal.
- A login already holding a challenge can answer it after the factor is deleted and mint a session that escapes any revocation.
- `_accepted` commits its recovery-code spend before `forget_factor` runs, so a failed deletion loses the code, and a concurrent enrolment can have proof from the old factor delete the new one.
- The offline command writes no audit event.
- `docs/self-hosting.md` tells people to remove the factor from Security settings. There is no such screen in the frontend.

Nothing here is worth salvaging piecemeal. The sequence is #427 first, then removal rebuilt on top of it with state checks, session revocation, an atomic verify-and-delete against a known factor id, and a test that runs the command rather than the function underneath it.

leonfullxr:
Superseded by #437, which rebuilds this from current main rather than patching it.

The premise this PR argued from was false: it justified removal costing a code by parity with enrolment, and enrolment cost nothing but a session. Three things had to land first, and each is now on main - #428 gave replacement a real bar, #432 split _accepted so proof is spent where the outcome is decided, and #434 made factor changes revoke sessions.

Every finding against this draft is addressed in #437, including the one that made it a draft: the command that built a coroutine, never ran it, and printed success. That is now driven through operator.main in a test, so reintroducing it fails.

One guard in #437 exists only because those landed in that order: removal spends the same attempt budget as replacement. Built first, as this PR was, removal would have been a second door onto the same six digits with no counter on it.

PR #428, opened 2026-08-25
Replacing a second factor costs the factor being replaced

### LINKED ISSUE #427: A stolen session with recent authentication can replace a second factor without knowing the old one
# Description

`POST /api/v1/account/totp` mints a fresh secret and hands it to the caller. `POST /api/v1/account/totp/confirm` verifies the submitted code against **that new secret** and then deletes whatever factor the account had, replacing it and its recovery codes.

Nothing in that sequence asks for a code from the factor the account already has. So a session with recent authentication can enrol a factor the session holder controls, in two requests, and the account's real second factor is gone.

# Motivation

`start_enrolment` carries a docstring explaining that it writes nothing until a code confirms, because writing first "made this endpoint the cheapest way to turn the second factor off: one request from a stolen session, no code, and no notice to anybody". That reasoning is sound and the deferral is worth having, but it protects against a factor being destroyed, not against one being replaced. The endpoint is still a way to disarm somebody's MFA from a stolen session: the attacker simply enrols their own.

The practical effect is that a second factor protects an account only until someone holds a session with recent authentication. Recent authentication means a password entered in the last thirty minutes, so an attacker who has phished the password and has a session is inside that window by definition. The factor stops them at the challenge, and then does not stop them at enrolment.

Nobody is notified. The account discovers it the next time it signs in and its authenticator no longer works.

# Functionality

Replacing an existing factor should require proof of the existing one: a current code from it or a recovery code, checked alongside the code proving the new secret. Enrolling for the first time has nothing to prove and should stay as it is.

The check belongs in `confirm_enrolment` rather than in `start_enrolment`, since starting an enrolment writes nothing and refusing it early would only tell an attacker to try a different route.

Worth deciding at the same time, and not implied by the above: whether a factor change should revoke the account's other sessions, which the specification says MFA events do and no code in `app/factors.py` does today.

# Details

- `backend/app/factors.py`, `start_enrolment`: seals a new secret and returns it, writing nothing.
- `backend/app/factors.py`, `confirm_enrolment`: `secret, codes = _opened_enrolment(...)` then `totp.matched_step(secret, body.code)`, where `secret` is the new one. The old factor is deleted immediately after, along with its recovery codes.

Found while reviewing #426, which added a removal route and justified its bar by saying removal should cost what enrolment costs. That justification does not hold while enrolment costs nothing but a session, so this is the more urgent of the two and #426 depends on it.

leonfullxr:
# Description

`POST /api/v1/account/totp/confirm` now requires proof of the factor it is replacing. Closes #427.

# Motivation

`start_enrolment` mints a fresh secret and hands it to the caller. `confirm_enrolment` verified the submitted code against **that new secret** and then deleted whatever factor the account held, along with its recovery codes. Nothing asked for a code from the factor being replaced.

Whoever asked for the new secret holds the new authenticator by definition, so that proof was worth nothing against the case it needs to stop. A session with recent authentication could enrol a factor it controlled and take the account's own out, in two requests, with nothing sent to anybody. The owner discovered it the next time they signed in and their authenticator no longer worked.

Recent authentication means a password entered in the last thirty minutes, and an attacker who has phished a password and opened a session is inside that window by definition. So a second factor stopped them at the challenge and then did not stop them here.

The endpoint's own docstring says enrolment writes nothing until a code confirms, because writing first "made this endpoint the cheapest way to turn the second factor off: one request from a stolen session, no code, and no notice to anybody". That reasoning is right and the deferral is worth having. It protects a factor from being destroyed. It never protected one from being replaced.

# Functionality

- Replacing an existing factor needs `current_code`: a code from the authenticator being retired, or one of its recovery codes.
- A first enrolment is asked for nothing, because requiring proof of a factor that does not exist would mean an account could never enrol.
- The delete is bound to the factor that was proved, and the request continues only if it is the one that removed it.

# Details

**The third guard was wrong before it was right, and the neuter check is what found it.** Binding the delete to the proved factor id was not caught by any test. Rather than write a test to cover it, working out *why* nothing caught it exposed a bug in the guard: binding the delete without checking that it matched is worse than not binding it at all. Two replacements racing would each delete a factor that was already gone and each insert their own, leaving the account holding two factors with nothing to say which one gates a login. So it became bind-and-verify, and it now has a test that holds both callers at a barrier and asserts one factor and one refusal.

**All three guards neuter-checked:**

| Removed | Failing test |
|---|---|
| the `current_code` requirement | `replacing_a_factor_needs_the_one_being_replaced`, `a_wrong_code_for_the_old_factor_changes_nothing` |
| checking that `current_code` verifies | `a_wrong_code_for_the_old_factor_changes_nothing` |
| the bind-and-verify delete | `two_replacements_at_once_leave_one_factor`, three runs, one failure each |

**The two cases that must keep working are tested rather than assumed.** Somebody moving to a new phone still has the old one in their hand, and somebody whose phone is gone has the recovery codes; a replacement proved by either succeeds, and the new secret is what ends up stored.

**Rejected alternatives are in the decision entry**, including asking for the old code when the enrolment *starts* rather than when it is confirmed, which only tells a caller to come back by another route and costs an honest person a code before they have set anything up, and notifying by mail instead, which is optional on a self-hosted install and is a record rather than a control.

`make verify` from the worktree with `PYTHONPATH` forced to it: 905 backend, 247 worker, frontend and CSP green.

**This unblocks #426**, currently a draft. That PR argued removal should cost a code because enrolment already did; the argument was false when it was written and is true once this lands.

**Not addressed here, and separate:** `app/factors.py` revokes no sessions on any factor change, though the specification says MFA events rotate them. That applies to enrolment as much as to replacement and predates both this and #426, so it wants its own change rather than a rider on this one.

## Summary by CodeRabbit

- **Security**
  - Replacing an existing authenticator now requires its current verification code or a recovery code.
  - Incorrect replacement attempts are rate-limited, while the existing authenticator remains unchanged.
  - Successful sign-in resets the replacement-attempt limit.
  - Concurrent enrollment and replacement requests are handled safely.
  - Suspended accounts cannot confirm new authenticator enrollment.

- **Documentation**
  - Clarified that first-time enrollment does not require a replacement code, while replacement does.


leonfullxr:
Moved to draft. Two P1s, and the first of them means this does not close the hole it was opened for.

**The replacement is still brute-forceable.** The sealed enrolment lives thirty minutes and failed `current_code` attempts are counted nowhere. The attacker holds the new secret, because they asked for it, so they can supply a valid new code on every try and grind `current_code` through the six-digit space until one lands. The ten-attempt budget belongs to the login challenge and does not reach here. So the gate this PR adds is a gate you can knock on a million times.

**A concurrent first enrolment walks past the check entirely.** Two confirmations can both read `replacing is None` before either writes. The first installs a factor, the second takes the account-wide delete path, removes it, and installs its own with no `current_code` at all. The decision has to be made inside the write transaction, not before it.

**And a claim I made in three places is false.** I justified the bind-and-verify delete by saying a race could leave the account holding two factors with nothing to say which gates a login. `auth_factors` carries `UniqueConstraint(user_id, kind)` with `kind` checked to `totp`, so two rows cannot exist. The guard is still worth having, because without it the last writer replaces a factor it never proved, but the reason I wrote for it in `docs/decisions.md`, in the code comment and in the test's docstring is not the reason. That is the third time this session I have argued from a property I did not check.

The threat description is overstated too: with a factor enrolled, a phished password alone yields a challenge rather than a session, so the case is a stolen session or one opened before the factor existed, not a phished password plus recent authentication.

**Also real, all confirmed:** a suspended account can confirm an enrolment it started while active, since `confirm_enrolment` takes `current_principal` and never checks state; `_accepted` commits the recovery-code spend before the replacement is certain, so a later failure loses the code; the race test's barrier sits outside the application so the interleaving it claims to force is not forced; and two tests pass without checking the thing they are named for.

What the reviewer confirmed clean is worth keeping: `raise REFUSED` inside `session.begin()` does roll back the preceding recovery-code delete, which is the property I was unsure of and tested directly this round. Cross-path replay between login and enrolment is blocked. Missing, empty and cross-account codes all fail.

Rework needed before this is a fix rather than a speed bump: count attempts against `current_code` and bound them, move the replacing decision inside the write transaction, require an active account, and correct the three places carrying the two-factor claim.

Issue #436, opened 2026-08-27
No credential change rotates the token of the session making it


leonfullxr:
# Description

Every credential change in this codebase revokes the account's *other* sessions and leaves the current one exactly as it was. A password change does it, an address change does it, linking and unlinking a provider do it, and enrolling a second factor does it since #434.

So a copy of the current session's token survives all of them. Whoever holds that copy keeps the session, and its realtime socket, through a password change made specifically to evict them.

# Motivation

The whole point of revoking other sessions during a credential change is to end access somebody else has. It works when the intruder signed in separately. It does nothing when they have a copy of the token belonging to the browser making the change, which is what a stolen cookie is.

The specification says these events "revoke or rotate affected sessions as appropriate". Revoking is implemented; rotating is not, anywhere.

# Functionality

Rotate the token of the session making the change: mint a new one for the same account, revoke the old, and set the new cookie on the response. The person making the change stays signed in and every copy of their previous token stops working.

This belongs in `sessions.revoke_others` or beside it, so that all five call sites get it at once rather than one route at a time.

The cost worth deciding on: a response lost in flight leaves the client holding a revoked token and signed out, where today it leaves them signed in. That is a real regression in a rare case, traded against closing a copied-token hole in a common one.

# Details

- `backend/app/sessions.py`: `revoke_others`, which revokes `Session.id != resolved.session.id` and leaves that one alone.
- Callers: `app/credentials.py` (password, email, identity link and unlink) and `app/factors.py` (enrolment).

Raised in review of #434. Deliberately not fixed there: rotating for factor changes alone would leave the codebase inconsistent, and a password change is the stronger case for it, so the decision belongs to all five call sites at once rather than to whichever one happened to be under review.

Issue #435, opened 2026-08-27
A login in flight can mint a session that bypasses the factor just enrolled


leonfullxr:
# Description

`confirm_enrolment` revokes the account's other sessions, but a login already in flight can create a new one immediately afterwards, and that session is not covered by the revocation.

The sequence: a login reads the account's factors, finds none, and is on its way to minting a session. Enrolment commits, and revokes every session that exists at that moment. The login then mints. The resulting session was never revoked, and its holder is inside an account whose owner has just turned on a second factor, past a gate that did not exist when the login started.

An old-factor challenge reaches the same place from the other direction: it can finish its proof, let a replacement revoke sessions, and mint afterwards.

# Motivation

Small window, and the reason it is worth a number is who is standing in it. Enrolling a factor is what somebody does about an intruder they suspect. An attacker who is signing in at that moment keeps a session the owner believes they have just ended, and nothing tells either of them.

It is not exploitable on demand: the attacker has to have a login in flight in the same instant. But it is the exact case the feature exists to close.

# Functionality

The login path needs to notice that the world changed under it. Two shapes:

- Re-read the factor after minting, inside the same transaction that mints, and refuse the session if one appeared. That keeps the check next to the thing it protects.
- Take the same per-account lock `begin_challenge` already uses, so a login and a factor change cannot interleave at all. That is a wider lock than the problem needs, and it puts a lock on the hot path of every sign-in.

The first is cheaper and more targeted. Either way the session must not be issued rather than being issued and then revoked, because a revoked-after-issue session is one the client already has a cookie for.

# Details

- `backend/app/accounts.py`, the login path: reads `factors.enrolled_factor`, then mints.
- `backend/app/factors.py`, `confirm_enrolment`: revokes inside the replacement transaction.
- `backend/app/factors.py`, `answer_challenge`: mints after the proof, and can mint after a concurrent replacement.

Found in review of #434, which added the revocation. The race is created by that PR only in the sense that there was nothing to race with before.

Issue #421, opened 2026-08-25
Concurrent logins each get a fresh TOTP guess budget, so the ten-attempt limit does not bind


leonfullxr:
# Description

`factors.begin_challenge` carries the account's spent attempts onto the new challenge, so that starting another login does not hand an attacker a fresh budget. It does that correctly one login at a time and not at all when the logins overlap.

Measured against the shipped code. A challenge with 7 attempts already spent, then four `begin_challenge` calls dispatched together, leaves these live challenges:

```
sequential, one after the other:  [7]
four concurrent:                  [0, 0, 7]
```

Two of the three live challenges start at zero. Repeat the pattern and the budget never runs out: whoever holds the password opens logins in parallel and gets ten guesses per challenge, as many times as they like.

# Motivation

Ten guesses per account per challenge is the whole of the brute-force limit on a second factor. A six digit code is one in a million, so an unbounded number of tries is the difference between a factor and a formality. The attacker already needs the password to reach a challenge at all, which is exactly the situation the second factor exists for.

The comment above the code states the intent plainly: "The budget belongs to the account, not to one challenge. Minted per challenge it counts nothing, because starting another is free to anyone who already has the password." That reasoning is right. The implementation does not achieve it under concurrency.

# Functionality

The carry-forward is a read-then-write across two statements with nothing serialising the account. Any of these closes it:

1. Take a transaction-scoped lock on the account before the update and insert, the way `SETUP_LOCK` already serialises setup-token minting.
2. Keep the budget on the account rather than on the token, so a counter is incremented rather than copied.
3. Allow one live challenge per account with a partial unique index, the way `asset_shares_one_active` allows one live share per asset, and let the loser of the race be told to retry.

The third has a precedent in this codebase and makes the invariant the database's business rather than the caller's.

# Details

`backend/app/factors.py`, `begin_challenge`: the `UPDATE ... RETURNING attempts` consumes the outstanding challenges and `sum(spent)` seeds the new row. Two transactions that both find nothing outstanding both seed zero. Under READ COMMITTED the second also seeds zero when the first has already consumed the row, because the re-evaluated `consumed_at IS NULL` no longer matches.

[...]

PR #437, opened 2026-08-27
Let an account take its second factor off, and the machine clear it

### LINKED ISSUE #419: A lost authenticator is unrecoverable: nothing removes a second factor, including the offline commands

leonfullxr:
# Description

`DELETE /api/v1/account/totp` removes a second factor at the same cost as replacing one, and `make auth-clear-factor` handles the account that can present neither a code nor an authenticator. Closes #419.

This replaces the draft in #426, which was built on a false premise and is superseded rather than continued.

# Motivation

Nothing removed a factor. Three routes existed and none of them turned one off; the only `delete(AuthFactor)` sat inside confirming a replacement. So "optional for every role", which the decision entry says and the code otherwise honours, was true of turning a factor on and not of turning it off.

An account whose authenticator broke could only spend recovery codes until they ran out, and was then unreachable. For the last administrator that was worse: `python -m app.recovery` prints a password link and never touches `auth_factors`, so the challenge still stood behind the new password. An install could reach a state recoverable only by `auth-collapse`, which destroys every account on it, and "every way back into a locked install is a command at the machine" was not true.

# Functionality

- `DELETE /api/v1/account/totp` takes recent authentication, an active account, and a code: a current one from the authenticator or a recovery code. The factor and its recovery codes go together, and the account's other sessions end with them.
- Guessing at removal spends the same attempt budget as guessing at replacement.
- `make auth-clear-factor EMAIL=...` removes a factor from the machine, ends every session the account had open, and records an audit row.
- A decision entry recording the bar and why a session alone does not meet it.

# Details

**#426 was abandoned rather than fixed, and this is what changed underneath it.** That draft argued removal should cost a code because enrolment already did. It did not: enrolment checked the submitted code against the *new* secret and deleted whatever the account held. Three things had to land first, and each is now on main:

- **#428** gave replacement a real bar, so parity with it means something.
- **#432** split `_accepted`, so proof is spent where the outcome is decided rather than where it is checked.
- **#434** made factor changes revoke sessions, so removal had somewhere to inherit that from.

**One guard exists only because those landed in that order.** Removal spends the replacement budget. Without it, removal would have been a second door onto the same six digits with no counter on it, straight past the budget #428 added. That is invisible if you build removal first.

**The bug that made #426 a draft is now caught rather than trusted.** Its `_run_clear_factor` called `_connected(clear_factor(email))`, but `_connected` is async and takes a callable, so the command built a coroutine, never ran it, and printed success while the factor stayed exactly where it was. The test called `operator.clear_factor` directly and never exercised the command. This test drives `operator.main(["clear-factor", ...])`, and reintroducing that exact bug fails it.

**Every guard neuter-checked:**

| Removed | Failing test |
|---|---|
| the code requirement | `removing_a_factor_costs_the_factor` |
| recent authentication | `removing_needs_recent_authentication` |
| the active-account check | `a_suspended_account_cannot_remove_its_factor` |
| the shared attempt budget | `guessing_at_removal_spends_the_same_budget_as_replacing` |
| session revocation | `removing_a_factor_signs_the_other_sessions_out` |
| `asyncio.run` on the command | `clear_factor_command_removes_it_and_ends_the_sessions` |
| session revocation in the command | the same |

**Traps hit while writing the tests**, all previously recorded and all worth the reminder: mixing the portal runner's event loop with a `TestClient`'s gives "attached to a different loop"; the operator command disposes its own connection, so anything read afterwards needs `db.connect(serving=False)`; and a scripted import rewrite mangled a multi-line `from app.tables import (` that already contained everything being added.

`make verify` from the worktree with `PYTHONPATH` forced to it: 924 backend, 247 worker, frontend and CSP, `MAKE_EXIT=0`.

**Not addressed here.** #435, a login in flight can mint a session past a factor change, applies to removal as much as to enrolment. #436, no credential change rotates the token of the session making it.

# Review fixes

Its own review found two P1s and three P2s. All are fixed here.

**A confirm whose authority died in flight could install a factor it controls.** `confirm_enrolment` decided whether to demand the old factor's code from a read taken outside its write transaction. A stolen session holds a sealed enrolment for thirty minutes and cannot confirm it while the account has a factor, because it has no code from that one. The owner taking the factor off is exactly what turns the same request into a first enrolment, which asks for no proof at all. Removal deletes the factor and revokes the other sessions in one transaction, so reading no factor means that commit has already happened, and a session read taken after it sees the revocation too. That ordering is the whole of the guard.

**The operator command could not close the sockets it said it closed.** `close_revoked` walks a dictionary that exists only inside the process holding the socket, and `make auth-clear-factor` runs in a process of its own where that map is empty. It revoked the session rows, closed nothing, and printed that it had ended every session the account had open, while the canvas kept drawing and kept its GPU slot. The process that owns the sockets now sweeps as well, asking PostgreSQL which of the account sessions behind them are still live. That is the source of truth for a session without Redis, so it is a periodic read rather than a second store, and it covers direct SQL and any later replica as much as it covers the commands. It fails open where the handshake fails closed, because the handshake judges one socket it is about to hand a slot to and the sweep judges every socket at once.

**One lock, where there was no row to lock.** The routes excluded each other with `SELECT ... FOR UPDATE` on the factor row, which holds nothing when the account has no factor yet. `clear-factor` and a first enrolment could therefore interleave: the command's delete of the factor finding none, the enrolment committing between that statement and the next, and the command then deleting the recovery codes the enrolment had just written. Every factor-changing transaction now takes a per-account advisory key first, before any row lock, which is what makes it safe to add to routes already arranged against deadlock: a lock every holder takes first can serialise them but can never be the second edge of a cycle. The row lock stays, because ordering `auth_factors` before `recovery_codes` is a separate invariant and still required.

**`clear-factor` on an account with no factor was not a no-op.** It deleted the recovery codes, ended every session, and recorded a `factor.cleared` row saying a second factor had been taken off an account that never had one, while printing that there was nothing to remove.

**Measured rather than argued.** The interleave above was reproduced on two connections before it was fixed: a delete that cannot see an uncommitted row, a commit, and a second delete in the same transaction that can, leaving one factor and no recovery codes. The exclusion and its release at commit were measured the same way.

**A bug on main that this uncovered.** `begin_challenge` has taken the same advisory lock since #421 and caught `OperationalError` around it. A lock timeout reaches SQLAlchemy as a plain `DBAPIError`, so that catch never fired and the 503 it documents has been a 500. Routing all four callers through one helper fixes it once; the test is parametrised over the three routes and the login path fails without it.

**Every fix neuter-checked:**

| Disabled | Failing test |
|---|---|
| the `is_live` re-check | `a_confirm_whose_session_died_in_flight_installs_nothing` |
| the same, made to refuse everyone | `removing_a_factor_and_enrolling_a_fresh_one_here_still_works` |
| the account lock in the two routes | `a_route_gives_up_rather_than_queue_on_the_account[first enrolment, removal]` |
| the account lock in the command | `clear_factor_waits_for_the_account_lock` |
| the wide `DBAPIError` catch | all three of `a_route_gives_up_rather_than_queue_on_the_account` |
| the no-op early return | `clear_factor_command_removes_it_and_ends_the_sessions` |
| the audit event alone | the same |
| the socket sweep | `a_session_revoked_in_another_process_is_swept_off_the_socket` |
| the sweep's live check | `a_sweep_leaves_a_live_session_drawing` |
| the sweep's error handling | `a_sweep_that_cannot_reach_the_database_keeps_live_sockets` |
| scheduling the sweep | `the_running_app_schedules_the_sweep` |

The last one is there because the other three call `close_dead_sessions` by hand: the single line in the lifespan that schedules it could be deleted with all three still green and no socket ever swept.

`make verify` from the worktree with `PYTHONPATH` forced to it: 938 backend, 247 worker. This branch changes no file under `frontend/`.

**Still not addressed here.** #435, a login in flight can mint a session past a factor change, which review found on removal as a third path and which belongs at the mint site rather than in any one route. #436, no credential change rotates the token of the session making it.

# Second review round

An independent pass over the five commits above measured all four claims they rest on and found no P1. Two findings, both fixed here, and one open question answered.

**The wiring test was still a level short.** It looked for a task by its coroutine name, so `await close_dead_sessions()` could be deleted from the loop body with all four sweep tests green: the other three call that function by hand, and the fourth never reached the loop. It now shortens the interval before the client starts and counts what the scheduled loop actually calls. Both neuters fail it: removing the `create_task` line, and removing the call inside the loop.

**A lock timeout outliving the wait it was meant to bound.** `SET LOCAL lock_timeout = '3s'` lasts for the whole transaction, and the `try` covers only the advisory statement, so a row lock taken further down timed out into a `DBAPIError` no handler expects and reached the caller as a 500. This is now lifted once the account key is held, which restores the wait those routes always did. `begin_challenge` had the same defect before this branch and gets the fix with them.

**The attempt budget was counting the wrong credential.** Removal and replacement charged one of ten tries before looking at what was presented. That is right for six digits, where a wrong guess must cost (#427), and wrong for a recovery code: twenty symbols from an alphabet of thirty-two is not a space anybody guesses at, so counting it bought nothing and cost the way back. Whoever stole a session could spend all ten on rubbish and leave the owner holding a working recovery code that no route would read, which is the case removal exists to serve. A presented value is now matched against the unspent codes first, in a session of its own so it takes no lock and cannot join the `auth_factors`-before-`recovery_codes` cycle, and only a value that is not one of them costs a try. Ten wrong guesses still shut the door on the authenticator's own code; the test asserts both halves.

**Corrections to what the comments claimed.** "Nothing has been written" is true of the write transaction and false of the request: an attempt counted by the previous transaction stays counted through a 503. The decision entry says so now.

| Disabled | Failing test |
|---|---|
| the call inside the sweep loop | `the_running_app_runs_the_sweep` |
| scheduling the sweep | the same |
| the recovery-code lookup before the charge | `a_spent_budget_does_not_take_the_recovery_code_with_it` |

`make verify` from the worktree with `PYTHONPATH` forced to it: 939 backend, 247 worker, frontend and CSP, `MAKE_EXIT=0`.

## Summary by CodeRabbit

- **New Features**
  - Added the ability to disable two-factor authentication using a valid authenticator or recovery code.
  - Added an offline recovery command for removing a lost second factor.
  - Factor removal deletes recovery codes and ends other active sessions.
  - Revoked sessions are automatically disconnected from live connections.

- **Security**
  - Requires recent authentication and enforces existing attempt limits.
  - Records an audit event for recovery-based factor removal.

- **Documentation**
  - Documented the new API endpoint and account recovery procedure.


leonfullxr:
Moved to draft. Three P1s, and one of them is a defect I shipped to main a few hours ago rather than anything this branch introduced.

**The offline command cannot do what it says.** `clear_factor` calls `sessions.close_sockets`, which walks `realtime.sessions`, an in-process dictionary. The command runs in its own process, so that dictionary is empty and the call does nothing. Database sessions are revoked; a live realtime socket keeps drawing and keeps its GPU access. `deletion.py` and `states.py` make the same call and theirs work, because those are HTTP routes running inside the API. Mine is the only caller outside it, and I copied the line without noticing the difference. My test checked database rows, so it proved the half that worked.

**A pending replacement can walk past a removal.** A confirmation that read the factor before removal, and resumes after it commits, sees no factor at all, takes the first-enrolment path, and installs its own factor having proved nothing and spent no budget. A stolen recent session can use that to install a factor it controls on an account that just removed one.

**A challenge can still mint a session after removal revokes them.** That is #435, filed against enrolment, and it applies here too.

**And #438, which is on main:** `answer_challenge` takes `recovery_codes` then `auth_factors`; `confirm_enrolment` takes them the other way round since #432. Those two deadlock, and a login answering with a recovery code is one half of it. That is my incomplete fix: #432's review said "use one lock order for both routes" and I applied it to one. This branch inherits the same cycle by taking `auth_factors` first, which is the right order, against a route that still takes the wrong one.

**Also real, all mine:** clearing a factor from an account that has none is not a no-op - it still deletes codes, revokes sessions, closes sockets and writes a `factor.cleared` audit row that did not happen; the `FOR UPDATE` locks nothing when there is no factor, so a first enrolment can slip between the factor delete and the code delete and be left with a factor and no recovery codes; and a stolen recent session can burn all ten removal attempts, which matters most for somebody down to their last recovery code.

Sequencing again: #438 wants fixing on main first, because this branch's lock order is correct and the route it deadlocks with is not. Then the removal race and the no-op behaviour here.

PR #476, opened 2026-09-07
refactor: share file cleanup for jobs and accounts

### LINKED ISSUE #474: Consolidate file cleanup for jobs and account deletion
# Description

Deepen the existing file cleanup function so jobs and account deletion use the same bounded-delete and failed-key recording rules.

# Motivation

Account deletion calls jobs._bounded_delete and jobs._trim_error, then repeats the delete-and-record sequence used by job cleanup. Both callers must know how one cleanup rule works.

# Functionality

Jobs and account purge use the same cleanup implementation. Each caller keeps its required commit order. Local and S3 storage continue to use the existing Storage interface.

# Details

- Generalize the existing _purge_keys function in backend/app/jobs.py into a public cleanup function with a caller context label. Delete deletion._forget_object and use the shared function for the account's keys.
- Keep retry sweep, task tracking and startup in their present locations. No new cleanup module file is needed.
- Keep attempt-key selection in the job module, and preserve cleanup after the job verdict commits.
- Preserve object cleanup before account row removal, indefinite retries, bounded backoff and the one-time alert.
- Preserve existing job tests for failed deletes, repeated recording, successful retry, cancellation and shutdown. Add account-purge tests with a failing storage adapter and cancellation during deletion.
- Sharing the existing cancellation shield improves account purge: cancellation records remaining keys for retry. This is an intentional behavior change.
- Preserve the existing best-effort behavior if PostgreSQL is unavailable; this refactor does not claim new crash durability.
- Run the backend gate and full make verify with the worktree's PYTHONPATH before push.

leonfullxr:
# Description

Use the existing blob purge function for account deletion and job cleanup.

Closes #474.

# Motivation

Account deletion called private job helpers and had its own object deletion loop. That loop did not record the remaining keys when cancellation stopped a purge.

# Functionality

A public purge function accepts the keys and a caller label. It handles bounded deletion, failed-delete records and cancellation. Account deletion now uses it, and its duplicate helper is gone.

# Details

Review the shared purge call and account row order, then the failed-storage and cancelled-purge tests. Retry tasks, tables, backoff and startup stay in place. Failed-delete recording remains best effort if the database is unavailable.

Checks: full make verify passes with the branch PYTHONPATH: 1010 backend tests, 297 worker tests and 106 frontend tests, plus type, lint, production build and CSP checks. Account deletion has 18 passing tests; job cleanup has 94.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved account deletion reliability when storage items cannot be removed immediately.
  - Failed deletions are now tracked and retried automatically, helping ensure cleanup completes successfully.
  - Cancelled account purges now preserve outstanding cleanup work so it can be recovered later.
  - Improved cleanup reporting with clearer descriptions of the items being processed.

### DIFF
--- a/backend/app/deletion.py
+++ b/backend/app/deletion.py
@@ -282,8 +282,7 @@ async def _purge(user_id: uuid.UUID, cutoff: datetime) -> None:
             select(Asset.storage_key).where(Asset.user_id == user_id))).scalars().all())
     # Before the rows: a key nothing names any more is an object nobody will
     # ever find, and the row is the only thing that names it.
-    for key in keys:
-        await _forget_object(key)
+    await jobs.purge_keys(keys, what=f"purged account {user_id}")
     async with db.session_factory() as session:
         async with session.begin():
             await hold_the_account(session, user_id)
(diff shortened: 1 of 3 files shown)


leonfullxr:
The docstring-coverage warning does not need a code change. The shared purge function documents the ordering rule, and tests cover the behavior. Adding routine docstrings to meet a coverage target would conflict with the repository comment rules.

